PR #4, opened 2015-02-26
Meta tags, title, OG, Cache Control in .htaccess

tradzik:


.htaccess
```diff
@@ -0,0 +1,8 @@
+<IfModule mod_headers.c>
+<FilesMatch "\.(jpg|jpeg|png|gif|swf|JPG)$">
+Header set Cache-Control "max-age=4838400, public"
+</FilesMatch>
+<FilesMatch "\.(css|js)$">
+Header set Cache-Control "max-age=4838400, private"
+</FilesMatch>
+</IfModule>
```
samdroid-apps: I added these rules in nginx, so I don't think we need this file.

Actually, does nginx even read the .htaccess files?

samdroid-apps:
Looks good!  Just could probably drop the htaccess.

davelab6:
I agree, looks good

.htaccess
davelab6: There's no harm in having that file included in the repo, right?

davelab6:
I suggest this can be merged.

davelab6:
Replaced with #40

PR #6, opened 2015-02-26
Please... navy blue is really ugly in this case.

ty221:
It is also very tiring for eyes. Tested on myself and my friends.


ty221:
@walterbender Maybe let us think about a background image...

PR #9, opened 2015-02-27
Give the website a sugar like design

samdroid-apps:


SAMdroid-apps:
Just pushed a change that cleans up the activities section

SAMdroid-apps:
I've put it online so you can see  https://activities-2.sugarlabs.org/www-sugarlabs/

PR #19, opened 2016-04-13
Create CNAME

davelab6:
Per https://help.github.com/articles/using-a-custom-domain-with-github-pages/ this should be added before our DNS servers are updated with the github server IPs

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,0 +1,1 @@
+sugarlabs.org


davelab6:
I suggest this can be merged; even if the DNS isn't reconfigured, it will do no harm.

samdroid-apps:
I suggest this could be closed.  It adds an extra file that is unused.  It e can't move to github pages, we need SSL (we use HSTS so it will break), legacy redirects, etc.

Issue #29, opened 2016-04-21
no school server mention


sdalybe:
Teachers and school IT admins will want to know about the school server offering available.

This could go under "What is Sugar Labs", a hyperlink with text
School Server Resources
and URL
http://schoolserver.org/

PR #31, opened 2016-04-21
Fix issue #29

erilyth:


davelab6:
I suggest this can be merged.

PR #58, opened 2016-12-07
Removed wikipedia image from the Google Graph

samswag:
Google doesn't work with Wikipedia image, so i uploaded our own image instead


tradzik:
It looks ok for me, won't merge as you have already merged it.

BTW It would have worked when Wikipedia link was `https://upload.wikimedia.org/wikipedia/en/6/67/Sugarlabs-logo.png`. However, link to our own resource is also fine.

quozl:
Link to our own resources is best; to take advantage of connection that already exists to our server.  Important for users with poor or bufferbloat connectivity.

PR #91, opened 2017-11-30
Fixed not working Sugar Labs promo video (fix #82)

### LINKED ISSUE #82: video with its code on line 51 of index.html does not play
The video with its code on line 51 does not play. this was tested on both chrome and firefox browsers.

a look on the  code shows that there is a problem with the placement of the  `#<video>` end tag.  
nice work sugarlabs!

maxibanki:
Referring to the Google Code-in task.


samswag:
Thanks, i see you figured out what we were ignoring. Is youtube placing any suggestions after the video is complete? is it possible for you so send me a video of this working on your computer? showing how it started playing and ended? let me see if it is smooth. Because the reason we used HTML 5 video player instead was due to these problems.

samswag:
Thanks, for the figuring it out. But i wouldn't merge yet because of the video, we have a task to create a promotional video, and we intend to feature that video on our homepage. But great job!!!!

IfeanyiEkperi:
Nice work @maxibanki

PR #107, opened 2017-12-01
Social icons update.

kiridix:


kiridix:
Changed ".instagram_colour" by ".ssk-instagram" that is the class of Instagram icon ( this add the gradient background to the icon ).
I have change the angle of the bottom gradient and put blue instead of purple in the top gradient.

Then added a background-color to ".ssk-github" that is Github´s icon´s class to add the black background to it.

Otherwise, I added a rounded corners to both icons.

walterbender:
It seems that the code you were stating from was a mix of minimized code with unminimized changes. It would make it much easier to read if you made a sequence of commits to unminimize the code, make your changes, and then minimize it again. Other than that, it looks good to me.

samswag:
Hi, i have a question. Why is it that the code only works for mobile version but not desktop. I am using a Macbook Pro running the latest google chrome. I did clear my cache.

Issue #114, opened 2017-12-01
broken link in https://sugarlabs.org/about-us/


capucinecoudert:
[Dextrose] points to [https://sugardextrose.org/] wich is broken.

yashagrawal3:
@walterbender should I just point it to Dextrose wiki page until broken URL  [https://sugardextrose.org/] 
 problem is not resolved?

quozl:
It should be removed, as it is no longer current.

Issue #115, opened 2017-12-01
Suggestion on page https://sugarlabs.org/about-us/


capucinecoudert:
"Harmonic distribution" and "Replacing Textbooks" are pointing to DEPRECATED and OBSOLETE pages, wich is confusing for the user.

quozl:
How are deprecated or obsolete pages confusing?  It seems quite clear what they mean.

These links should be removed, as the projects that they linked to are no longer current.

PR #122, opened 2017-12-02
Fix issue #114 : broken link in about-us.html

yashagrawal3:
@quozl  please review.


quozl:
Yes, looks like it will fix the issue, but I've a couple of comments;

- [ ] #115 could be fixed at the same time, saving reviewer time,

- [ ] best not to mention issue numbers in first line of commit message (as we might not always use GitHub, but would still use Git), mention issues in body of commit message, see [Making commits](https://developer.sugarlabs.org/contributing.md.html#making%20commits) for further information.

PR #142, opened 2017-12-02
Icon update on the 'Get Sugar' section

infinitelooped:
I updated the color and the design of icons on the 'Get Sugar' section. Currently on it's second version, Here's what it looks like:

icon v2

Feedbacks are appreciated. Thank you


pro-panda:
Stale. The current design looks the same to me

Issue #121, opened 2017-12-02
CSS should minify on build


kipply:
Having to commit CSS is a pain for the version control as well as prone to errors. The CSS should minify while building to speed up the development process, make version control easier and to prevent errors with the most up to date minified-file.

amaaniqbal:
We are not using minified CSS files anymore.

PR #159, opened 2017-12-04
added #d465d6 color on hover on links

### LINKED ISSUE #81: Improve link colors
### The links inside purple sections
These are hard to see and the hover color is even harder to read. I suggest making these links be  `#fa99e3` and `#d465d6` (the color in the logo) during hover. They look really nice after testing.

### The footer links
The normal color is good but the hover color isn't. I suggest making the hover color for these links be `#d465d6` (the color in the logo). It looks really nice after testing it.

sanatankc:
Closes #81 

### FIxed links hover color:

screen shot 2017-12-04 at 11 30 04 am


quozl:
Reviewed.  Why does the minified css change look so different?  Is it a different minifier?

sanatankc:
Maybe, I don't know which minifier, we suposed to use?

quozl:
I've no idea, so I'll speculate.

Perhaps the same minifier you used in your previous commit https://github.com/sanatankc/www-sugarlabs/commit/fa8b48277c6b60b4eb91dd600f1ec002d3ef6b0f or we might ask others what they used when they changed the same file recently; @samswag, @kipply, @eohomegrownapps.

sanatankc:
I just search on internet, use whatever I find. @kipply has mentioned before, we should have a task runner like gulp or grunt which handles all our minifying, compressing related tasks automatically on every file changes.

quozl:
Yes, we should have common steps.  I've not been able to find where it was mentioned as an issue.  If you agree we don't have it listed as an issue, please make one.

I've reviewed your changes including c6f264b and am not sure you aren't introducing some additional change to the minified css.  I saw a spacing change as I scrolled it right.  You might use a simplistic beautifier on both versions of the minified css to compare it.

sanatankc:
I think last time I committed wrong one. This one looks fine to me.

sanatankc:
In #121 , @kipply mentions about minifiying css

quozl:
Thanks.  Looks fine now.

PR #109, opened 2017-12-01
Fix the unclickable slider

kipply:
Fixes #85 

For some reason, this slider wasn't using `owl` like all the other ones. I left the slider nav left aligned in this version, since I thought all the centered text looked a bit off. Thoughts? 

Also most of the line changes are just removing trailing whitespaces, sorry for those of you who are looking at the diff!

PR #160, opened 2017-12-04
Fix broken navigation

### LINKED ISSUE #158: Navbar broken
navbar broken if viewport width is smaller than 991px and bigger than 768px on desktop, tested on chrome ubuntu

images

kipply:
When working in #109 I accidentally deleted a bracket which broke some CSS. Sorry for the error!

Adding that bracket back fixes #158


quozl:
Thanks.  Saw something while reviewing; your minified file makes an additional change; from `background:url(../assets/bg6.jpg` to `../assets/new_bg6.jpg`, please either;

- [ ] explain why the same change is not made to the original file, or;

- [ ] remove the change and put it in another pull request.

kipply:
@quozl That wasn't an intentional change. Not sure how it got in there, though it may have something to do with the minifier or the fact that the file can't be located. That's been fixed, thanks!

quozl:
Thanks.  Verified with wdiff.

```
wdiff -3 old new
 [-(max-width:1199px){.boxDownload2{height:855px}@media-] {+(max-width:1199px){.boxDownload2{height:855px}}@media+}
```

Issue #173, opened 2017-12-08
bad path for background image in minified css


walterbender:
In airspace.css, the background image is specified as ../assets/bg6.jpg
When that css file is minified, the minified file is put into the minified subdirectory.

This breaks the relative path resolution.

404 file not found: www.sugarlabs.org/css/assets/bg6.jpg

Tested on Chrome and Firefox on Fedora 27

Seems there are three approaches we could take:

(1) move the minified files into css, giving them a new name, e.g., airspace.minified.css (This would require a change to _layouts/base.html, where the minified css is accessed.);

(2) modify the path in the minified code (there are actually four occurrences of relative path names that probably should be changed) but this would probably lead to long-term maintenance issues as the expectation is that minified files are not changed from their originals except in format;

(3) add a symlink or copy the asset directory into css (This seems both clumsy, error prone, and obscure, again presenting long-term maintenance issues.)

I would argue that the best choice forward is no. 1 above.

There maybe other options. And, reportedly the background image somehow works on some version of Safari, although I cannot image how it could work.

walterbender:
FWIW, in https://www.google.com/url?q=https%3A%2F%2Fgithub.com%2Fgeekrypter%2FsugarLabsWebsiteRedesign&sa=D&sntz=1&usg=AFQjCNH2SMJn3nUELs2QxgXbOPB2BRYBHA option 2 was used: the relative path in the minified file is ../../assets/

quozl:
Stale, closing.  We have since stopped minifying our derivative copy of airspace.css.

PR #175, opened 2017-12-08
Rename airspace.css to airspace.minified.css

Rohith-coder:


walterbender:
Is this in regard to #173 ? How does renaming this file solve the problem?

Rohith-coder:
yes it is regarding #173  and I followed the instructions given so how should the issue be solved?

walterbender:
There were 3 steps in the instructions: rename, move it to the css directory, and update _layouts/base.html. You only did one step.

PR #177, opened 2017-12-09
New Search Bar

AdithyaKrishnaK:


quozl:
- [ ] the commit messages do not explain what you are changing or why, so they are difficult to review and will make for difficult reading when people are reviewing commits that have been merged.  Please see [making commits](https://developer.sugarlabs.org/contributing.md.html#making%20commits) for guidance on how to write commit messages,
- [ ] the change to minified css seems very large, as if you are using a different minifier to everyone else, and there is no change to the non-minified css, please check again that your change is intended and minimal.

AdithyaKrishnaK:
The changes are  in the beginning and in one of @media only screen and (max-width:767px) in airspace.css

quozl:
Thanks for the commits.  Everybody watching Sugar Labs on GitHub received a mail message for every commit you made.  The commit messages were not useful, despite my earlier review comment that gave guidance.  Please consider;

- [ ] using a local clone on your own computer instead of GitHub's editor, so that we don't all see your every edit,
- [ ] using more meaningful commit messages; which form part of a self-reflective practice.

Thanks!

PR #180, opened 2017-12-09
Fixed Search Bar Responsiveness

samswag:
I made some minor changes to solve the problem with responsiveness,
whenever the browser window shrinks.
screen shot 2017-12-09 at 11 02 09 pm


walterbender:
This is not really about responsiveness. It is about a problem with elements overlapping when the window shrinks. By moving the search box slightly to the left, it no longer overlaps with the About entry in the top menu as the window shrinks.

AdithyaKrishnaK:
Please take a look at pull request 182

PR #182, opened 2017-12-11
Search bar optimisation

AdithyaKrishnaK:
This is the make the search bar responsive. The changes are to nav.html and airspace.minified.css . The changes makes sure that search bar moves with the navigation bar. This is done by getting position search bar and navigation bar and adjusting the position of search bar. This also removes animation from mobile view. Changes to airspace.minifed.css is to correct the position of search icon.

### DIFF
--- a/_includes/nav.html
+++ b/_includes/nav.html
@@ -1,9 +1,22 @@
 <!-- Header for screens with width > 1260px and width <= 767px -->
 <script>
 if(window.innerWidth > 1260){
+	var expanded = false;
 	$(document).ready(function(){
+		$("#SearchBar").focus(function(event){
+			event.stopPropagation();
+			expander();
+		});
+		$(document).click( function(){
+			if(expanded){
+    			condenser();}
+		});
+		$("#searchForm").click( function(event){
+			if(expanded){
+			event.stopPropagation();}
+		});
 		var searchbar = $("#searchForm").offset();
-		var menubar = $("#menuBar").offset();var distance1 = menubar.left -(searchbar.left + 200) ;	 
+		var menubar = $("#menuBar").offset();var distance1 = menubar.left -(searchbar.left + 200);	 
 		if((searchbar.left +200)>menubar.left){
 			var distance= distance1 + 13 + searchbar.left;
 			$("#searchForm").css("left", distance.toString() + "px");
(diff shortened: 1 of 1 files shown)


walterbender:
I am having a hard time reviewing this since I don't know what problems the various patches are addressing (please update the commit messages on each separate commit with an explanation of what is being accomplished by the commit).

Also, I am a bit confused by the term "responsiveness". Is that what you really are addressing? Will these changes make the search bar respond more quickly? If so, it is not obvious when looking at the commits.

Finally, making changes directly to the minified code is bad practice as the next time airspace.css is minified, your changes will be lost. Please make them to the unminfied file and then minify.

AdithyaKrishnaK:
# ### 17cd472 - The search button looks a little odd on mobile view. This is because the input is not wide enough. I have made increased the width of input.
c368d92 - Eliminating animation from mobile view
1fd78ec -

AdithyaKrishnaK:
17cd472 - Search button was oddly positioned in mobile view. Fixed increase in width of input.
c368d92 and 1fd78ec - To remove animation from mobile view
aa578b8 - By responsiveness I mean that the search bar will move along the navigation bar rather than ticking to a spot. The code for it is the jquery one on the top.

quozl:
I agree with @walterbender.  My earlier review comments in #177 for @AdithyaKrishnaK were missed, so I'll patiently say it again differently.

We need well-written commit messages for several reasons;
- for reviewers to understand your change in the pull request before it is merged,
- for problem solvers to understand your change in the git commits once merged,
- for deep problem solving using `git bisect`, `git blame` and the other advanced tools.

It isn't good enough to explain the commits in the pull request as you have here, because the pull request is not saved in git.  It is only held in GitHub.  git will never show your explanation.

So, please;
- [ ] read our check list for [making commits](https://developer.sugarlabs.org/contributing.md.html#making%20commits), and if you have any questions please ask,
- [ ] learn to use a local clone rather than GitHub's web-based editor,
- [ ] rewrite your commits in this pull request (using `git rebase -i` or other tools) and then force push.

Thanks.

p.s. you've also made these commits and pull request on your `master` branch, which goes against [our conventions](https://developer.sugarlabs.org/contributing.md.html#branching); it prevents you from working on more than one set of changes at a time.  In future, please think about branching.  It isn't clear if you intended 27d32a9 for this pull request.

AdithyaKrishnaK:
Yes, 27d32a9  is included.

quozl:
You're still using `master` branch, I see.

PR #186, opened 2017-12-18
Search bar improvisation

AdithyaKrishnaK:
- Making the search bar 'responsive' by making it change it position with respect to navigation bar.
- Removing animation from mobile view.
- Closing or reducing search bar size to normal on clicking anywhere else.
- improvements in looks of search bar.


quozl:
@walterbender, it was said you wanted the minified file unminified, edited, and minified, and so I'll presume you have some way to diff it and review.  When I use my diff tools, there's a lot of different.

samswag:
Can you send a photos or gifs showing the changes?

AdithyaKrishnaK:
You can get preview here https://adithyakrishnak.github.io/

walterbender:
@quozl  As has been discussed in several threads, we need a uniform process for handling minification. It should be that the unminified file should be the primary place we make edits. If we decide on one tool-chain for minifying, we should be OK. I have no opinion about what that tool should be.

quozl:
@walterbender, thanks.

At the moment, GitHub does not render minified files in a line by line fashion suitable for change review.  Best I've come up with is to fetch the user branch and do;

```
git diff --color-words origin/master -- css/airspace.minified.css
```

If the chosen minify tool could also use new-lines instead of spaces every now and then, we'd be able to review more easily without leaving GitHub. :grin:

@AdithyaKrishnaK, tested, thanks.  But the word diff of your minified file shows extensive changes unrelated to your change.  I'm ready to merge your pull request if either;
* the minified file doesn't change in more ways than the unminified file (i.e. use the same minifier as other people have used), or;
* the pull request doesn't change the minified file (i.e. remove that change and force push).

PR #187, opened 2017-12-18
Update .htaccess - Enable Compression and Leverage Brower Caching

prezine:
The Added .htaccess Codes 
1. Enable Compression, and
2. Leverage Browser Caching


quozl:
Call for reviewers.  Does this really improve performance by lowering data demand, or is our server already using compression?

PR #190, opened 2017-12-23
Added temporary solution to minify CSS and JS

### LINKED ISSUE #189: An automate process using gulp
This has been mentioned before.
@kipply mentioned it in #121,
@quozl also mentioned in a comment in #159 

It is hard to make changes in current setup. First we have to make change in source file then minify it is using a minifer. Everyone is using different minifier.

It would be really good if we could have some sort of process using gulp (or grunt).

kipply:
Here is a temporary solution to minifying CSS and JS. 

By running `grunt`, CSS and JS will automatically be minified as updates to the unminified files are made. Ideally, minified files won't have to be committed at all, and minification can occur server-side. This will at least allow people to make updates more easily, and keep consistent minification. 

Fixes #189


.gitignore
```diff
@@ -1,3 +1,4 @@
 _site
 .sass-cache
 .jekyll-metadata
+node_modules/
```
quozl: Missing newline at end of file.

package.json
```diff
@@ -0,0 +1,28 @@
...
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/sugarlabs/www-sugarlabs"
+  },
+  "author": "",
+  "license": "ISC",
+  "bugs": {
+    "url": "https://github.com/sugarlabs/www-sugarlabsissues"
```
quozl: Missing slash before issues.

package.json
```diff
@@ -0,0 +1,28 @@
...
+    "grunt-contrib-cssmin": "^2.2.1",
+    "grunt-contrib-uglify": "^3.2.1",
+    "grunt-contrib-watch": "^1.0.0"
+  },
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/sugarlabs/www-sugarlabs"
```
quozl: I'm not sure what this phrase is for, but many of use use git+ssh instead of git+https.

README.md
```diff
@@ -43,7 +43,7 @@ If some changes are not visible (after saving a page while editing), please rest
 ## Important points to be noted
 
 1. Whenever the code is changed, please ensure that the relevant code changes are reflected in **main.js** file (For example, if a slideshow is disabled, make sure the id of the slideshow element/section is also disabled in the **main.js** file if it exists).
-2. Whenever a new commit is made and the changes are made to any one of airspace.css, main.js, or plugins.js files, try to use a minification service and add the minified version of new code in respective **minified** folder. (css and js folders have a **minified** folder)
+2. HTML files read from minified CSS and JS files. While making changes to unminified files, you should use `grunt` to watch changes to the unminified files, and grunt will automatically minify them as needed. Remember to `npm install` before install necessary dependencies before running `grunt`.
```
quozl: Imperative phrasing needed?  After making changes to some files, use grunt to update the minified file.

quozl:
Also, could you remove the minified files from the repository and have the server generate them during deployment hook?  @scanterog, what do you think?

kipply:
If the server just added a step that ran the grunt process, then we could add minified files to the .gitignore. This would fix a lot of conflicts, and having to update PRs when multiple peolpe are working on CSS or JS.

quozl:
Yes, but I've no access, so we have to wait for Sugar Labs systems@ team to respond.  @walterbender, suggestions?

walterbender:
Has anyone contacted @scanterog ?

scanterog:
Hi guys! Sorry to be late to the party. Just provide the steps needed to add to deployment script and I will add it.

quozl:
@scanterog, any progress?  Or can you tell me what I should do?  I've established necessary access to the servers.  I don't know the mechanism by which a commit to master will update the web server filesystem.  With GCI 2018 started, we're starting to be hit with pull requests that include minified files, again.

kipply:
Added a package lock that controls more versions than `package.json`

Nepaltechguy2:
Would using Jekyll to straight up serve minified versions of files like the ideas [here](https://stackoverflow.com/questions/10208481/how-to-compress-and-minify-assets-using-jekyll-plugins) work? 

I was just thinking of other possible ideas, and this might simplify the setup process.

kipply:
@quozl Based on that thread, there are two ways we can do this without changing anything on the server; 
 - Wrapping the jekyll build in an npm function that'll also execute minification (this would be nice as we can also add linting at this level) 
 - Creating a simple jekyll plugin that executes it with jekyll hooks

Nepaltechguy2:
I'm on the boat of adding Node/grunt/gulp to this, I just wanted to know the reasoning behind not writing a plugin.

quozl:
Do we actually need to use minified files?  While it is a convention with larger sites with extreme demand to optimise as far as possible, this optimisation is costing us more developer time than it is worth.  We don't have that much demand.

Our server should support gzip compression, and our use of OpenSSL may be using run-length encoding.

Would someone like to do a performance characterisation with and without minified files?  One way to do this is with a virtual machine configured with a restricted CPU bandwidth.

Nepaltechguy2:
Doesn't minifying code only make it smaller for the client to download? In my fork of the repo, the size of the minified file is 26% smaller than the non-minified.

quozl:
The relative size of the two files is unimportant.  It is such a tiny file.  After gzip compression, the file is 9682 bytes long, or 7 network packets at 1500 bytes each.  Yet loading the page costs 4.96 MB over 79 requests, over three thousand network packets.  Why are we even _bothering_ to minify a file this small?  Habit?

Other two minified files are main.js and plugins.js.  I'll make a pull request for you each to review.

Nepaltechguy2:
Whoops, I didn't even look at the file sizes :( Agreed, minifying that tiny of a file is causing far more problems than it is worth.

quozl:
https://github.com/sugarlabs/www-sugarlabs/pull/248

PR #198, opened 2018-02-26
Fixed search not working in all pages

abdulsamad234:
I changed line 42 of nav.html so that when the search is pressed, the request that is sent holds the general search url which is https://sugarlabs.org//search?q=... because it seems like the server side code is not setup to respond to search requests from other pages.

### DIFF
--- a/_includes/nav.html
+++ b/_includes/nav.html
@@ -39,12 +39,12 @@
     }
 	function searchLoad(){
 		var re = document.getElementById("SearchBar").value;
-		document.getElementById("searchLink").href = "https://sugarlabs.org/"+ "/search/?q=" + re;
+		document.getElementById("searchLink").href = window.location.hostname + "/search/?q=" + re;
 	};
 	document.addEventListener('keyup', function(ev) {
 		if( ev.keyCode == 13 ) {
 			var re = document.getElementById("SearchBar").value;
-			document.getElementById("searchLink").href = window.location.href + "/search/?q=" + re;
+			document.getElementById("searchLink").href = window.location.hostname + "/search/?q=" + re;
 			document.getElementById("searchLink").click();}
 	});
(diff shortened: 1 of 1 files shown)


abdulsamad234:
As the previous method will only work when the website is hosted with sugar labs. I changed the `window.location.href,` which will return the whole location url, to `window.location.hostname` which only returns the url in this format: www.example.com

quozl:
What happened to the required URL prefix `https://`?  It was in `.href` before, right?

abdulsamad234:
Yes, i could easily concatenate it back, but i thought it would work without it since it'll like redirect to it automatically.

quozl:
But will it redirect to http or https?  Is the redirect browser-specific?

samswag:
@abdulsamad234 I have tested the code locally, I don't think it address the issue here maybe from my side. How did you test it? 
it is looking like this http://127.0.0.1:4000/turtle-blocks-js/127.0.0.1/search/?q=sasmos 

I think the problem of the search working outside the homepage is because of this www-sugarlabs/search.html. The "Search" only works when it in this order "https://sugarlabs.org//search/?q=sugar" and not when it is "https://sugarlabs.org/turtle-blocks-js//search/?q=sugarlabs" so the aim should be eliminating the search and embed it somewhere within "nav".

@geekrypter ?? Thoughts?

geekrypter:
Instead of using ```window.location.href + "/search/?q=" + re ``` or ```window.location.hostname + "/search/?q=" + re```, we could use ```"{{ site.baseurl }}/search/?q=" + re``` since the baseurl is captured through the config.yml file. I have tested this method on my machine and it worked.

quozl:
@geekrypter, thanks!

geekrypter:
@quozl, you are welcome :)

PR #200, opened 2018-03-06
suggested security update

walterbender:
There is a vulnerability reported in nokogiri and it is recommended to update to v 1.8.1

Gemfile.lock has been updated to:
nokogiri ~> 1.8.1

Not 100% sure how to test this.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -62,7 +62,7 @@ GEM
       listen (= 3.0.6)
       mercenary (~> 0.3)
       minima (= 2.0.0)
-      nokogiri (= 1.6.8.1)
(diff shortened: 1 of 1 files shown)


samswag:
I saw the security update but wasn't sure how I can test it either.

Let me explore the internet or maybe @geekrypter should have an answer?

quozl:
I saw the vulnerability weeks ago and decided I would never post about it in public. :grin:

quozl:
Was it meant to be `-` or `~`?

geekrypter:
@samswag, I think your latest commit is the solution to the issue.

Issue #171, opened 2017-12-06
Improve of side social buttons style - suggestion [GCI]


TheLukaszNs:
I think that these buttons could be styled better - for example adding border-top-[**left/right**]-radius to only first element and border-bottom-[**left/right**]-radius to last one would be much better instead of having one without border radius, next with it and so on.
This: ![gci2](https://user-images.githubusercontent.com/25536472/33676887-47002748-dab7-11e7-9926-f9d9ae8f3bad.png) can look like this: ![gci2 1](https://user-images.githubusercontent.com/25536472/33677227-586d8042-dab8-11e7-98ed-daf4adc5df47.png)

Oluebube:
Am a new contributor can i work on this?

kushagra189:
@TheLukaszNs  is this still up or already assigned?

quozl:
* mostly nobody needs permission to work on anything, just announce if you are working on it, so that your time is not wasted by someone else working on it at the same time,
* asking permission may give no response, as has already happened for @Oluebube, and that doesn't mean much,
* making a pull request to fix a problem that doesn't have wide consensus can stimulate a discussion to achieve that consensus, and that discussion may fail, which then wastes your time.
* it is best to have wide consensus before working on problems of style and design,
* in my opinion, this issue is of no consequence; don't mind if it is fixed, and don't mind if it is not fixed.

TheLukaszNs:
Yeah, as @quozl said - **I'm not the one that allows you to make PR that will resolve this issue**. You can work on any issue you want - it's up to you if you want to contribute!
<hr />
I saw that this issue isn't resolved so far, so go on - do it if You want 😄

kanishkarj:
Hey ! I would like to work on that

quozl:
With ec6074b merged, my opinion is that it looks uglier than before; the eye is drawn to the disparity of the straight browser edge and the unnecessary curve away from the edge.  What do the regular contributors think?

kanishkarj:
How does this look? I tried to cover up that disparity
![screenshot from 2018-03-13 09-35-55](https://user-images.githubusercontent.com/22411349/37322203-3004ab22-26a2-11e8-9134-1f9015c9b826.png)

Pro-Panda:
I agree with @quozl 
I removed the browser edge curve from the social buttons in 06f282c. Does this look better ?
Also, 
[...]

PR #203, opened 2018-03-11
Improve of side social buttons style - suggestion [GCI] #171

kanishkarj:
As per the discussion on pr #171 , I styled the social media buttons. this is how it looks now :
screenshot from 2018-03-11 19-28-59


quozl:
I've reviewed the change but I do not understand it.

* [ ] there are changes to trailing whitespace in base.html; are these necessary or accidental?
* [ ] social-share-kit.css was changed from one line to many, making it impossible to see the actual changes (if any) in git diff shown by github,
* [ ] a new minified file is added, which we've already seen in recent pull requests is adding to maintenance costs, re #189.

kanishkarj:
@quozl Ill address each of these issues one by one:

- [ ] they are not necessary, they are accidental. I have changed the Stylesheet link in base.html to the minified file, maybe while doing that I left some whitespaces.

- [ ] The thing is social-share-kit.css is a minified CSS file by itself. I don't find it wise in editing a minified CSS file. As CSS by itself is hard to understand and minifying it makes it worse. Hence I made a separate minified file so that the changes can be made to social-share-kit.css and then minified to social-share-kit.min.css.  

- [ ] Yes I have seen that issue before. I think its more efficient to have a minified file. And I can write a proper gulp process to minify files appropriately, in fact, I was working on it without realizing someone has done that.  

Please let me know if you want me to edit the minified file itself, or if you want me to leave it like this and write a gulp process.

quozl:
Thanks.  But as nothing about the change has changed since eb3c034, I still don't understand it.  Is it necessary to change the file between minified and not-minified in order to fix the problem?

kanishkarj:
Actually, it's not required. I can put back just the minified file. But think of it like this, every time someone wants to edit the styles file they will have to beautify the code, make changes and then minify it back.
But what if we could make changes and the gulp process takes care of the rest. easier right!

the changes can be seen in the GUI, notice the sidebar : 
before:
screenshot from 2018-03-12 12-24-19

After:
screenshot from 2018-03-11 19-28-59

quozl:
What did you change?

kanishkarj:
I changed CSS. and the stylesheet path in HTML to the minified file. in CSS i added styles for ssk-facebook and the rest

quozl:
But I can't see the change in the commits.  I can only see the files after you've made multiple changes.  So my review is blocked.

kanishkarj:
So what changes am I supposed to do?

quozl:
You've made too many changes at once, so reviewers cannot see easily and exactly which change fixed the problem #171.

You might split your commit into several.

kanishkarj:
Ahh, shall work on that. Thank you

kanishkarj:
Hey! Ill close this Pull request and open a cleaner one. Thanks

PR #204, opened 2018-03-12
Improve of side social buttons style - suggestion [GCI] #171

kanishkarj:
I have tried a cleaner implementation of it.


quozl:
Thanks.  Your change can be seen using the side by side view of the commit ff339e2.

add

However, the commit message does not meet our [convention](https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#making-commits) so I'll rewrite that for you.

Issue #149, opened 2017-12-03
Mobile View Issue+ Suggested Solution


MrMalfunction:
![image](https://user-images.githubusercontent.com/24508987/33523075-7b9c3c58-d823-11e7-898c-6b330799925f.png)
That's a lot of space user might wait for something to appear there.
Solution decrease the width between to the two elements.

samswag:
Thanks for the update

MrMalfunction:
I will keep Updating.

quozl:
Is also an issue on Firefox-57 on desktop, so I'll reopen.

avinashbharti97:
@quozl  @samswag  @MrMalfunction 

should it look like this?


![after_padding_change](https://user-images.githubusercontent.com/22724583/37400725-d1235846-27ab-11e8-82e5-5a6eb1343bc8.png)

quozl:
In general, vertical distance should be consistent between the elements.

In particular for this issue, it is the distance from the three dots down to the "Have a taste".

In comparison, consider the distance;
* from the "Try now!" down to the background colour change.
* from the background colour change down to "GET SUGAR",
* from each background colour change on the page up or down to the next element.

Is also an issue on Firefox-58.0.4 on desktop.

quozl:
@avinashbharti97, https://github.com/sugarlabs/www-sugarlabs/pull/209 didn't seem to make any difference for Firefox on desktop.

avinashbharti97:
@quozl changes were reflected in firefox when i was checking on local server, However i will check again

quozl:
Yes, I checked on a freshly installed laptop with no cache.  Perhaps the distance has reduced a little bit.

avinashbharti97:
okay i will check again, thanks for information

PR #209, opened 2018-03-22
mobile view padding of try sugar on web set to 70px top and 70px bottom

avinashbharti97:
padding_fixed

**In mobile view padding is set to 70px top and bottom which is even to other elements padding** #149


quozl:
Missing change to `css/airspace.css`?

Having to change two files; https://github.com/sugarlabs/www-sugarlabs/issues/189

PR #210, opened 2018-03-26
Added Projects Thumbnails  #183

### LINKED ISSUE #183: Projects Thumbnails Suggestion
In the about us page, in the projects section, it would be more expressive if we can have thumbnails or cards types of projects description. Somewhat like in the gif below.

divi-thumbnail-border

VatsalBatra:
replaced simple links with thumbnails in project section of about section 

fixed #183  
 
sugarlabs-about


quozl:
You have committed a file `.DS_Store` which tells me you did a `git add` of everything instead of carefully adding each change.

See also https://github.com/sugarlabs/www-sugarlabs/issues/183 for my comment there.

VatsalBatra:
@quozl  thanks for guidance , can you specify what do you mean by more genric logos as the logos I used  for  Gsoc and Soas are the commonly used one  , also should i make the necessary changes as stated by  you or leave this  issue for the time and wait

quozl:
It is really up to you.  Use your independent thought.  If you think a logo is needed, and a logo is not available, then your only choices are to either use some generic logo or ask the project team for one.

VatsalBatra:
So , I have removed the obsolete projects from the list  and have corrected the 'alt' tag .Also, the heading and the links in the projects are accessible for those who use screen reader, so adding thumbnails won't create any issue. Any other suggestions are welcome...

quozl:
Thanks, reviewed.

VatsalBatra:
@quozl  I don't have the write access ,can you please guide me how to merge it into main branch .

quozl:
@VatsalBatra, thanks for the patch, I've reviewed it.  As they have been involved in https://github.com/sugarlabs/www-sugarlabs/issues/183 the other people who might review are @lalitmee, @manan-goel-1999, and @samswag.

For merging, I'm not willing to do that, because, as I said in https://github.com/sugarlabs/www-sugarlabs/issues/183, "in my opinion, the section does not need logos".  Consensus is not clear.

You'll have to wait for someone to be willing to agree with @lalitmee that "it would be more expressive" and therefore should be merged.  At least you've done the editing and testing.

Looking at it from my understanding of the point of view of each of the others;
- @lalitmee could say "hey, that's not how my screenshot does it",
- @manan-goel-1999 could say "hey, my idea was slightly different but i didn't get time to do it", and;
- @samswag could say "hey, i don't care, just merge it."  :grin:

samswag:
@quozl I did review. I like the thumbnail feature, but the search bar is having some conflict. @VatsalBatra could you look into this, or better see if you can get a better search bar position.
screen shot 2018-03-30 at 10 21 18 pm

quozl:
@samswag, thanks.  @VatsalBatra you have made changes to many files, not just `about-us.html`.  Can you reproduce the search bar change @samswag showed us above?  If so, please update the pull request to avoid the change.

VatsalBatra:
@samswag thanks for pointing out ,I have corrected the searchbar issue.
sugar_labs_about
@quozl and  @samswag please review.

samswag:
Saw it, I think you should update the repo on your local machine. The background on the homepage stop displaying.
screen shot 2018-03-31 at 10 32 01 pm

lalitmee:
@VatsalBatra I think if we can add a little border on the thumbnails it would be a little bit easy to recognize and it would look good. :smile: 
Btw nice efforts. :+1:

VatsalBatra:
@samswag   done.

@lalitmee thanks, I tried that but it was not looking good, may be because no adjacent component has   border but I think we can add it in the future if when number of projects increase.

lalitmee:
@VatsalBatra actually I am saying that if you can do something like [this](https://user-images.githubusercontent.com/10762218/33958481-85ced90c-e06a-11e7-92e2-375d157af809.gif) so that it is easy to distinguish between all for projects. Actually in current view its like we can't really tell after seeing the project that where one is starting and where it is ending i.e borderline. :nerd_face:

quozl:

PR #211, opened 2018-04-03
extra padding removed from all responsive view

avinashbharti97:
#149


kipply:
The cascading properties of CSS should probably be used here instead of `!important` tags. It'll make styling/debugging more difficult in the future as it breaks the natural CSS cascading properties. 

https://developer.mozilla.org/en-US/docs/Learn/CSS/Introduction_to_CSS/Cascade_and_inheritance

quozl:

PR #213, opened 2018-04-22
 #issue 174 -> add colours to activities icon

kushmisra:
added colours to activities icon in their original SVG format.


quozl:
Thanks.  Using [guide for reviewers](https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#guide-for-reviewers).  Reviewed a0ecb70.

* [ ] neither your commit nor your pull request are linked to the issue, so I'll write here; probably fixes #174, there that should do it, :grin:
* [ ] your commit message doesn't explain the problem or how it was solved, see [making commits](https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#making-commits) for some suggestions,
* [ ] the SVGs are base64 inline embedded, so using GitHub alone I cannot review the impact of your patch; I would have to run up an instance of the web site; which I don't have time to do right now, so next time you might add a screenshot, 
* [ ] after downloading the changed index.html and opening in browser, the colours are not what Sugar looks like; this is a critical failure.

For me it is important that that the web site represent what Sugar looks like, and you didn't notice the difference.  You should run up an instance of Sugar Live Build and learn how Sugar works.  Read the documentation on https://help.sugarlabs.org/

The issue of colours was raised in https://github.com/sugarlabs/www-sugarlabs/issues/174, please review that discussion.

Will you fix colours and push further changes?

PR #235, opened 2018-10-18
Remove extra # from airspace.css and airspace.minified.css

### LINKED ISSUE #234: Bug in CSS code
When I try to minify the airspace.css using gulp minifier i.e., `gulp styles`,  it throws the following error.

capture

The bug is in line 378. It has unnecessary # before class is defined.
capture

amaaniqbal:
Fixes #234 
This commit removes extra # from airspace.css and airspace.minified.css which is preventing gulp minifier from minifying airspace.css


quozl:

PR #236, opened 2018-10-18
Add gulp minifier to automate css and js minifying process

### LINKED ISSUE #189: An automate process using gulp

amaaniqbal:
Fix #189 
This commit adds gulp minifer to automate css and js minifying process.

To minify airspace.css file just execute `gulp styles` from root directory and to minify main.js & plugins.js simultaneously, execute `gulp scripts` from the root directory. The resultant files will be stored in the required location as it is currently.

Prerequisite is that you must have [https://nodejs.org/en/](https://nodejs.org/en/) installed. (8.12.0 LTS used here).


samswag:
Is it possible to add the instructions on how to use the automation on the Readme.md? I am away from my computer? how do I know this is working? is there any screenshot I can see? @quozl want to look into this too?

amaaniqbal:
This is how it works
capture

You just need to install [node js](https://nodejs.org/en/) and run required gulp commands to do the minification.  Or if readme is required, where should I add it. Should I modify the main one?

quozl:
Thanks.  Looked at briefly, but it looks impractical to review.  Please consider;

* [ ] you have changed thousands of files in this commit; please confirm that every single change is required, (e.g. by splitting the commit into several), or check that you haven't done a recursive `git add` by accident,
* [ ] include edits in your branch to the README so that people know how to use the new tooling,
* [ ] remove all minified files,
* [ ] ensure the commits explain each change.

Ask questions if you have any concerns.

amaaniqbal:
What do you mean by "remove all minified files" @quozl ?

amaaniqbal:
Mainly only one file i.e., gulpfile.js is required for minification purpose alongwith package.json . But then user has to do `npm install` to install all the dependencies specified in the package.json file. User may get tons of deprecated warnings and might have to find few additional fixes to them depending upon the version of node js they are using.

But in this pull request I did `npm install` using node js version 8.12.0 LTS and so all the required dependencies required for the minification purpose is installed. You can find them inside node_modules folder. And so user only has to run required gulp commands after installing node js version 8.12.0 to do required minification. This adds thousands of files inside node_modules and make the minification process version specific so that different PR wouldn't lead to conflict.

Kindly share your opinions regarding whether this PR should include node_modules before hand or should be left to the contributor to install them @walterbender @samswag @quozl

quozl:
> What do you mean by "remove all minified files" @quozl ?

As in https://github.com/sugarlabs/www-sugarlabs/issues/189 and the referenced https://github.com/sugarlabs/www-sugarlabs/issues/121 and https://github.com/sugarlabs/www-sugarlabs/pull/159, we would like to avoid pull requests which change minified files.  Minified files should be generated instead.  Thus they should be removed once they can be automatically generated.

quozl:
Ideal situation is that the minified files are;
* not present in the git repository,
* not edited or added to patches by contributors,
* not part of pull requests,
* added to .gitignore,
* created within contributor clones when required for local testing,
* created on the server by the infrastructure hooks that publish changes after commits are made, and;
* are not used except where there is a performance advantage in doing so.

amaaniqbal:
@quozl have you tried testing it locally?

And for resolving conflicts, can you suggest the best way to take these three commits and put on top of master? Naive approach would be to revert all three commits, rebase and apply the patches one by one.

quozl:
> @quozl have you tried testing it locally?

No, sorry.

> And for resolving conflicts, can you suggest the best way to take these three commits and put on top of master? Naive approach would be to revert all three commits, rebase and apply the patches one by one.

So many different methods, it is difficult to choose.  If you've not done rebasing before, read GitHub or Atlassian documentation on it, and practice with a backup of your repository clone.  It also depends on the nature of the conflicts.  One method is `git rebase -i`.  Another is `git pull --rebase`.  Another is to `git reset` and recreate commits.  Another is to start a new branch based on sugarlabs/master, `git cherry-pick -n` your changes, then `git add --interactive` (for which I use emacs magit), and create new commits.  There are other methods too.  They all end with a `git push --force`, which you can skip if you're not happy with how the branch looks in `gitk` or another visualiser.

kipply:
See #190  for reference. #190 does the same thing but with grunt(a bit more compact) instead of gulp. Only one of these pull requests will be necessary. 

Please do not commit `node_modules`!

quozl:
@kipply, thanks.  My preference is #190.

quozl:
https://github.com/sugarlabs/www-sugarlabs/pull/248

PR #253, opened 2018-11-08
Sugar Labs website: Modify navbar styling

SzymonLeja:
I think i made it properly :)
Please comment if u have any suggestions

### DIFF
--- a/css/airspace.css
+++ b/css/airspace.css
@@ -1487,7 +1487,7 @@ px      68    160    252    */
 
   }
   header.small{
-    height: 500px;
+    height: px;
   }
   header.small #navh{ width: 120px; height: 7px; margin-top: 14px; }
 .divider{
--- a/css/airspace.css
+++ b/css/airspace.css
@@ -1,4 +1,3 @@
-
 /*--
   Common Css
 --*/
@@ -154,6 +153,9 @@ header {
 header .navbar {
   margin-bottom: 0px;
   border: 0px;
+}
+#firstHeader.small .navbar-default{
+  box-shadow: 0px 1px 20px 0px;
 }
 header .navbar-brand {
   padding-top: 0px;
@@ -1485,7 +1487,7 @@ px      68    160    252    */
 
   }
   header.small{
-    height: 23px;
+    height: 500px;
   }
   header.small #navh{ width: 120px; height: 7px; margin-top: 14px; }
 .divider{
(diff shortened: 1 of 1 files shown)


quozl:
Thanks.  I wasn't able to review; not enough information.  More information needed in commit message or pull request; what was the problem to be solved, how was the problem solved, what testing done.  See https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#making-commits for ideas.

amaaniqbal:
The exact task was as follows :

"Mostly our website has a very light background(almost white) and our navbar also has a white background. When the page is scrolled, the exact navbar component area is almost invisible.

Add box-shadow to the navbar such that it appears to be floating, distinct and attractive. An example of such a navbar could be found [here](https://in.udacity.com/)."

quozl:
Thanks.  Wasn't that problem fixed in https://github.com/sugarlabs/www-sugarlabs/pull/240?  Can I see before and after screenshot?

amaaniqbal:
It was not any problem. The task just says to make current navbar appear floating and distinctive from rest other components especially for those part of pages where the background is almost white.

amaaniqbal:
@SzymonLeja add a vendor prefix too atleast `-webkit-`. And ensure that box-shadow is present on small and medium sized devices.

Also gradient is current a bit dark. Use 5th argument of box-shadow to give a very light color to the background in rgba format.

quozl:
Thanks.  I disagree with the task; it was probably not widely discussed.  I find the navbar easily recognised before and after scrolling.  So my vote is no.  I'll wait for consensus from the marketing team.  @samswag?

samswag:
@quozl  I tested it isn't solving many problems. I think if the color of the navbar can be changed into something visible compared to the white, then I think that will be a significant change.

For now, I agree with @quozl 
> Thanks. I disagree with the task; it was probably not widely discussed. I find the navbar easily recognised before and after scrolling. So my vote is no. I'll wait for consensus from the marketing team. @samswag?

SzymonLeja:
I just changed the color of the shadow and added webkits. What you think about it?

quozl:
@samswag, thanks.

@SzymonLeja, thanks for your work on this.  I've no time right now to spin up a test instance to see the effect of your change c47368b, but am happy to review a screenshot.

SzymonLeja:
gci
Is it okey?

quozl:
Thanks, I see a drop shadow underneath the navbar.

quozl:
Why close?

quozl:
Thanks.  As a practice, that will make me less willing to review.  No return on the investment of my time.

PR #254, opened 2018-11-09
Sugar Labs website: Make subscribe button consistent 

SzymonLeja:
It's GCI task.

"Currently subscribe button located in the footer has a default bootstrap behavior and rest all buttons have a customized behavior.

Update the subscribe button class to match those of rest of the buttons on our website. Also modify its padding using id such that the button does not look bigger than that of the above email input field."

I changed this button from <input> to <button> and set its class to the same as other buttons on site, i deleted width argument from its ID to make it smaller.

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -112,7 +112,7 @@
                                                     <div style="position: absolute; left: -5000px;" aria-hidden="true"><input type="text" name="b_23c6df6118693ee743df3e0a4_11cb7e5c96" tabindex="-1" value=""></div>
                                                     <div class="clear">
                                                         <div class='align-center-md'>
-                                                            <input type="submit" value="Subscribe" name="subscribe" id="mc-embedded-subscribe" class="btn btn-primary">
+                                                            <button type="submit" id="mc-embedded-subscribe" class="btn bt-raised btn-custom bt">Subscribe</button>
                                                         </div>
                                                     </div>
                                                 </div>
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -112,7 +112,7 @@
                                                     <div style="position: absolute; left: -5000px;" aria-hidden="true"><input type="text" name="b_23c6df6118693ee743df3e0a4_11cb7e5c96" tabindex="-1" value=""></div>
                                                     <div class="clear">
                                                         <div class='align-center-md'>
-                                                            <button type="submit" id="mc-embedded-subscribe" class="btn bt-raised btn-custom bt">Subscribe</button>
+                                                            <input type="submit" value="Subscribe" name="subscribe" id="mc-embedded-subscribe" class="btn bt-raised btn-custom bt">
                                                         </div>
                                                     </div>
                                                 </div>
--- a/css/airspace.css
+++ b/css/airspace.css
@@ -322,6 +322,8 @@ font header .navbar-default .navbar-nav li a:hover {
   margin-top: 30px;
 }
 .btn{border:0px solid transparent !important; /* this was 1px earlier */}
+#mc-embedded-subscribe{ width: 58% !important;padding: 6px;}
+
 /*--
   Feature Start
 --*/
--- a/css/airspace.css
+++ b/css/airspace.css
@@ -321,7 +321,6 @@ font header .navbar-default .navbar-nav li a:hover {
   color: #333333;
   margin-top: 30px;
 }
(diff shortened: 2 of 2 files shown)


amaaniqbal:
Only few minor modification required :
1. Don't remove the name attribute. It may be doing something in interacting with the backend.
2. Try not to change input to button tag. But nothing to worry here.
3. Align the button towards center on larger devices
4. See that the height of subscribe button is not larger than that of input field above it. If its there, just decrease its top and bottom padding a bit.

SzymonLeja:
Can button on mobile be bigger?

amaaniqbal:
Looks good to me. Actually, the button being larger in smaller devices is a bug which is present previously too. 

A possible solution for this could be to make the input email field as the same size as that of the button which currently looks most effective to me. This can be done by overriding the width inside using `!important` inside `#mc_embed_signup input.email` styling.

What do you say @quozl @samswag ?

quozl:
I'm aware that `!important` is a tool of last resort, and it is usually better to fix the underlying problem.  Other than that, I think the susbcribe button looks fine to me; you'd have to demonstrate by screenshot why it isn't fine.

quozl:
> Kindly go through the task @quozl.

Can't.  Tasks are private to GCI students and mentors.  I'm not a mentor.   Don't forget that Sugar Labs contributors aren't all mentors.

> Subscribe button has a completely different appearance than all other customized buttons on our website and looks like an odd one out.

I disagree.  There are no other buttons.  There are boxed links, but they aren't buttons, as they don't submit data.  The subscribe or join button is unique in this, and so it should appear unique.

Previous related pull requests are;
* https://github.com/sugarlabs/www-sugarlabs/pull/243
* https://github.com/sugarlabs/www-sugarlabs/pull/244

This seems to be needless churn.

I call for consensus on this issue first.  Let's have a discussion; is the button fine as it is, or does it need to change?  How does it need to change?

If the goal is just to provide a task for GCI, then fine, accept the task attempt from the student but don't merge the change.

amaaniqbal:
> Don't forget that Sugar Labs contributors aren't all mentors.

Ya, that's there, I meant Szymon already mentioned in the first comment exactly what the task is about.

> Previous related pull requests

That task was created by me only. At that time, its color was changing from sky blue to purple on hover.

> I call for consensus on this issue first.

I agree with you.

> If the goal is just to provide a task for GCI, then fine, accept the task attempt from the student but don't merge the change.

Okay, I will approve his task. Should this PR instead focus on resolving the following issue?
capture

quozl:
@amaaniqbal, do you mean a subscribe button wider than the text entry?  Reproduced with iOS 12.1 in portrait orientation.  Does it still happen after https://github.com/sugarlabs/www-sugarlabs/pull/258?  I feel silly worrying about such a trivial part of the web site that as far as I can tell doesn't work (what newsletter?), while we have Sugar activities that don't work, and new and fascinating ways in which Sugar isn't working.

amaaniqbal:
> @amaaniqbal, do you mean a subscribe button wider than the text entry?

Yes

> Does it still happen after #258?

Specifying width has nothing to do with the version of Bootstrap.

> I feel silly worrying about such a trivial part of the web site that as far as I can tell doesn't work (what newsletter?)

Sorry, will keep in mind from the next time.

quozl:
Is there any news on this?

amaaniqbal:
Not sure what should be done with it. Better to fix it.

As you said

> I feel silly worrying about such a trivial part of the web site that as far as I can tell doesn't work (what newsletter?)

If it doesn't work i.e., subscribers don't get any emails related to Sugar activities, then we should think of either removing it and redesigning the footer or make it functioning and keep our subscribers updated. 

What do you say?

quozl:
Thanks.  That the newsletter doesn't exist and doesn't need subscription is covered in https://github.com/sugarlabs/www-sugarlabs/issues/249 so it doesn't need to be redundantly tracked here.  It does need fixing, but it is a separate problem.

This pull request by @SzymonLeja is what I'm asking about; is the button going to be made more consistent or not?  Is @amaaniqbal going to do it instead?  Or has everyone given up and we can close the pull request?

amaaniqbal:
Yes, I can fix it. Along with the problem with the width of the button on some devices. But only after everyone's consensus. I feel this should be done.
This PR can be closed for now.

PR #258, opened 2018-11-13
Added Bootstrap4.1.x

Joeloedinho:
Added suitable approach to achieve Bootstrap 4.1.x


quozl:
More needed in commit message to explain what is changed and why.

There's a `css/.DS_Store` extra file that may reveal things about your computer; not sure if you want this in the commit.

Joeloedinho:
ok @quozl  i will do just that

quozl:
Was replaced with https://github.com/sugarlabs/www-sugarlabs/pull/259

Issue #247, opened 2018-10-26
Remove duplication between README.md and documentation.md


quozl:
Both files contain text similarities, leading to increased cost of maintenance.

Swayamsvk:
I think documentation.md file is deleted.

Issue #262, opened 2018-11-19
Purpose of all .sass files?


HarrisonLKatz:
Hello,
I am currently working on this website, and could not figure out the purpose of the sass files. I added an import to main, but discovered that main.css is never used on the website. When I imported it, it broke the website. Not only that, but one file, _social.sass, isn't even linked in main.sass. Is there a reason that these are kept, or should I ignore/delete them?
Thank you,
Harrison

quozl:
Thanks.  Interesting.  [SASS](https://en.wikipedia.org/wiki/Sass_(stylesheet_language)) is used by [Jekyll](https://en.wikipedia.org/wiki/Jekyll_(software)).  If someone has changed generated files, i.e. output files that are made from those `.sass` files, then;

* that's all-our fault for letting it happen, because not everyone would have known how Jekyll works, and;
* you can find from `git blame` and other views of history in GitHub why they did it, and ask them.

HarrisonLKatz:
No one has changed the output - I'm just confused because the output is never used. In fact, one of the files does not even output anything. I was wondering if they have any real purpose because I am trying to add some css, and it seems that main.sass outputs main.css which is never imported or used anywhere.

quozl:
Remove the unused files, test that the site still works for you, and make a pull request.  Then anyone who cares about the files will be obliged to give reasons to keep them.

HarrisonLKatz:
Resolved by #263

PR #263, opened 2018-11-19
Remove redundancies (GCI)

HarrisonLKatz:
Hello,

Sorry that this pull request took a little while to write - I got stuck on figuring out how some of the site's css worked. This pull request does everything I noted on my GCI task and also resolves issues #247 and #262.

Here is a quick summary of my GCI submission.
Since syntax.css is not used, I have removed it. Also, html5shiv was included inside modernizr, but was also called separately. I removed it from modernizr as calling it seperate is recommended; however, this had the side effect of updating modernizr. ~~I added the google verification codes to git ignore since they are only needed for the sugarlabs.org server and are not necessary for anyone else.~~ I also removed social share kit and replaced it with ion icons since the rest of the website was also using ion icons. This change did slightly change how the icons look, but not significantly, and it allowed me to remove the entire social kit library. I added the css to a seperate file, but since I based it off social kit I kept social kit in the license. 

I also removed all the .sass files since they are no longer used (#262). Lastly, I combined README.md and documentation.md (#247)

This should remove quite a few code redundancies and make it simpler for new people to start contributing towards www-sugarlabs, especially if anyone needs to change the social media icons.

For reference, this PR is based off my task submission for Google Code-In 2018. My mentors are @walterbender @amaaniqbal and @i5o, however my task was already approved since making a PR was not required.

Thank you,
Harrison


quozl:
Thanks.  Reviewed.  Especially like how the commits are broken out by purpose and each has an effective message.

I've not yet found what updates the filesystem on the server, but it will need changing before bbea49e is merged.

HarrisonLKatz:
Hello,

Thank you @quozl for reviewing this and for the compliment on my commits - I tried to break up my commits in order to clearly show why I removing these files.

Unfortunately, I probably can't help much with updating the server. However, from the limited information I could gather from wiki.sugarlabs.org/go/Service/www, it would seem that there would need to be a folder created somewhere where the google verification files could be added, and the python app would need to be updated with the following:
```python
from shutil import copyfile
# ...
# put this after the git pull...
copyfile('PATH_TO_GOOGLE_VERIFY_DIR/google5f7220d9ae752fa5.html', 'PATH_TO_WEBSITE/google5f7220d9ae752fa5.html')
copyfile('PATH_TO_GOOGLE_VERIFY_DIR/google6c0d01bc1a53fa32.html', 'PATH_TO_WEBSITE/google6c0d01bc1a53fa32.html')
# but before run jekyll build
# ...
```
However, I rarely use flask and don't know the actual code the website uses, so the above could be completely wrong. If there are any other issues with this pull request besides the server, or if you would like me to resubmit this PR without commit bbea49e, feel free to let me know.

Thank you,
Harrison

quozl:
Thanks, very useful.  I've just now learned _docker_, located the container, uploaded the definition to a repository [sugarlabs-infra/www-rebuilder](https://github.com/sugarlabs-infra/www-rebuilder).  There isn't an obvious (to me) place to add in these files, but anything is possible.

Help me understand though; the files removed by bbea49e are only useful to Google, grant access to Google Search data for individual Sugar Labs members, and are of no use to anyone else unless they want to grant the same access to those same members.  If so, why remove them?

Being able to accept new files of the same nature, without involving our sysadmins, would be lower cost.

HarrisonLKatz:
The main reason is explained here https://stackoverflow.com/a/23378145. However, I did not know how much extra work it would create - it would probably be simpler to remove that commit. I can't remove it right now but I will later today.

quozl:
Thanks.  Reviewed.  @samswag, please review.

samswag:
Everything looks fine to me. But @HarrisonLKatz did you address the issue with the Google verifications? those two files are for me and @davelab6 , we use it for capturing of numbers and related stuff.

quozl:
Yes, those files aren't touched any longer by the commits in this pull request.

samswag:
@quozl want to merge?

PR #277, opened 2018-12-01
Fixing Bug in airspace.css

Aniket21mathur:
Fix issue #234


quozl:
Thanks, but we already have a pull request.
https://github.com/sugarlabs/www-sugarlabs/pull/235

PR #280, opened 2018-12-09
Update index.html

aperezbios:


aperezbios:
Merging in changes with recommendations from @davelab6 incorporated

PR #282, opened 2018-12-12
Remove unused CSS

amaaniqbal:
This commit removed some unused CSS as pointed in a GCI task.

### DIFF
--- a/css/airspace.css
+++ b/css/airspace.css
@@ -377,28 +377,11 @@ font header .navbar-default .navbar-nav li a:hover {
   width: 120px;
   height: 120px;
 }
-#h2ele {
-  line-height: 27px;
-  color: #5C5C5C;
-  padding-top: 110px;
-}
-#testimonial .block h2 {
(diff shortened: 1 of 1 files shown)


quozl:
As _airspace.css_ is from an upstream project, there's good reason to stay synchronised, and not enough reason to desynchronise further.  The amount saved is trivial compared to the size of the cold cache page load.  So my vote is no.

amaaniqbal:
I agree with you. It's best to stay synchronized with the upstream unless there is a strong reason to desynchronize. I just wish to send the work of a GCI student. Also, there are three instances of this task and only two was completed. One was this and the other was #263.

quozl:
I'm not surprised.  GCI tasks are often hastily prepared and not reviewed.  Often it is only when the pull request arrives that actual review takes place.  Ignore GCI, and concentrate on the legitimacy of the proposed change?

amaaniqbal:
Then I think this should be closed.

PR #287, opened 2018-12-22
Adding Try button to MB page and updating screenshots of MB for carousel

### LINKED ISSUE #285: Proposal for link to sugar labs hosted Music Blocks.
The first result when searching in Google "Music Blocks" is the Music Blocks about page on the Sugar Website. (https://www.sugarlabs.org/music-blocks/). For young students, this can be confusing when they see the instructions on setting up a local web server when really all they need to do is navigate to https://musicblocks.sugarlabs.org which is not mentioned on the page.

I propose we add a prominent go-to button at the top to the Sugar Labs hosted release to make it more accessible to students, parents and teachers.

FreddieN:
This PR updates the pictures on https://www.sugarlabs.org/music-blocks/ and adds a try it now button so it users searching for the music blocks software can find it easier.
demo
Fixes #285


walterbender:
I think these assets make sense for Turtle Blocks, but less so Music Blocks. Maybe somethings like:

music-tree
hot-cross-buns-graphics
musical nautilus
crab-canon-plot
???

FreddieN:
@walterbender I agree, I have updated with more Music Blocks orientated projects. I have also added links so the user can view them inside Music Blocks.

walterbender:
What is the story with the CSS file? Why did that have to change?

FreddieN:
@quozl Took me a bit longer than a minute but I think it should be ok now...
@walterbender I added some styling to the try it now button which is in the CSS file.

quozl:
Thanks.  Reviewed.

* [ ] file `mb3.png` content is shifted down slightly.

Otherwise there's no other obvious problems, so I'll merge.

PR #291, opened 2019-02-23
Improve search bar responsiveness

manangoel99:
The issue was fixed by adding the navbar-form class from bootstrap to #SearchForm1 and #SearchForm divs of the _includes/nav.html file


amaaniqbal:
I tried this locally as well on the website directly. The added classes only change the vertical position of the search bar to something worse aligned.

@quozl you can close this.

quozl:
Thanks for the review.

Issue #163, opened 2017-12-04
About Page - Moving Cards


g30r93g:
The cards start moving before the user scrolls to see them. They should only start moving when the user has scrolled to that part of the page. A user may experience this if they were 
too slow when moving down.
![33514750-3fd22eda-d751-11e7-8643-75b4f2cf3fd9](https://user-images.githubusercontent.com/31825917/33567988-d40f0880-d91c-11e7-974f-ed4cd506ac3c.jpg)

akshat-vashistha:
can I still contribute on this issue @g30r93g ? I would be more than interested to work on it.

chimosky:
@akshat-vashistha please make your changes and open a PR if you have a fix. Thanks!

quozl:
This would be better forwarded upstream, and may have already been fixed there so an update of the theme may help.  Still, the work is not planned, so I'll close the issue for the moment.

PR #292, opened 2019-03-01
About page card fix

manangoel99:
This pull request is in reference to #163 .  It fixes the carousel moving only after the user has moved to the carousel.


chimosky:
Tested but it doesn't work as expected, it still behaves like it does on master.

PR #295, opened 2019-04-07
free/libre != open source

walterbender:
We refer to Sugar as an open source platform, but whereas we use a GNU license, Free/Libre is a more accurate description.


samswag:
Fine by me

PR #302, opened 2019-10-02
Remove alpha mention on Sugarizer

llaske:
Remove alpha mention on Sugarizer, it's no longer an alpha product :-)

### DIFF
--- a/index.html
+++ b/index.html
@@ -168,7 +168,7 @@ title: Sugar Labs
             </div>
             <div class="col-md-4">
                 <div class="border-box boxDownload">
-                    <h3 class="customHeaderStyle">Sugarizer<br>(alpha)</h3>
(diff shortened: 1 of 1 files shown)


llaske:
@quozl what's the process to update the website when a PR is merged?
Look like this PR was merged long time ago and still not online.

quozl:
It was automated.  On 6th November I wrote to systems@ (after @tchx84 reported 88bc4f1 had not shown up either), that starting the docker container `freedom:/srv/containers/org.sugarlabs.www-rebuilder` did not seem to work.  I don't know enough about docker yet to figure out what next to check, but my first guess is that operating system upgrade may be connected.

llaske:
@walterbender could you help here? We talked about that to the SLOB meeting yesterday.

walterbender:
I will try to look into it.

Issue #288, opened 2019-01-06
Search bar not very responsive across different screens


chukundah:
<img width="697" alt="screen shot 2019-01-06 at 3 24 06 pm" src="https://user-images.githubusercontent.com/14887641/50737346-43321d80-11c8-11e9-8d66-990485f3711d.png">

This is a 1280px width screen, and the search bar covers a menu item.

manangoel99:
Can I work on this issue?

pro-panda:
Please note there is no need to ask permission to work on an issue. See our [guide to contributing](https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#contributing)

manangoel99:
I have made a pull request. It has been labelled as issue #291

manangoel99:
This also fixes issue #214

akshgpt7:
@chukundah Is the issue still open?

quozl:
We're waiting for review of https://github.com/sugarlabs/www-sugarlabs/pull/291.

chukundah:
> @chukundah Is the issue still open?

It is.

quozl:
Please test and review https://github.com/sugarlabs/www-sugarlabs/pull/291

amaaniqbal:
I don't see any method to go about this by changing the margin, padding, or media-screen attributes. Even if somehow it gets fixed now, there would be some other silly error like this later. This was also happening in the past.

The best way I find is to replace this custom search box with the Bootstrap 3 default search box and add the classes accordingly for the search functionality.

deejoe:
Are the steps to reproduce this available? In Firefox 77.0.1 or Chromium 83.0.4103.61 on Ubuntu 18.04 I don't see the problem. Narrowing the width of the browser window or increasing the size of the application font using `Ctrl +` eventually leads to the the menu items to collapse into a "hamburger" menu (ie, three horizontal lines stacked atop one another) but never to the overlap of the elements shown in the original post. Might test on other device, browser, OS combinations as opportunity allows.

In the meantime, if someone can speak to whether another change might have fixed this, or that the bug still exists (and if so, under what conditions) that could be helpful. 

Also, to clarify, #291 was closed on request of the submitted, who tested it and reported that the result of applying the patch was not acceptable.

Might as well close this if no one can reproduce the bug nor has a patch, is my feeling.

quozl:
I agree.  Please close.

deejoe:
[...]

PR #303, opened 2019-10-05
Fixed search bar overlapping

akshgpt7:
The search bar is not overlapping with other navbar items.


akshgpt7:
Issue #288

quozl:
Thanks.  Please test and review https://github.com/sugarlabs/www-sugarlabs/pull/291 as the change there is much smaller than yours, in terms of number of files and lines changed.

Issue #289, opened 2019-01-11
Include a back-to-top button on scrolling down


rbhatia46:
On scrolling the website, a back-to-top button can be included in the bottom-right corner to make it easier to navigate back to the top.
Please let me know if you need any help with this or if I can integrate this in the existing UI.

quozl:
Thanks.  We need the content to be verified and updated.  We need the time on page to be increased. Layout and appearance is lower priority and not as necessary as content.  So I'm not in favour of what you suggest.

rbhatia46:
@quozl thanks for the reply! I was looking forward to make my first contributions to Sugar Labs. Can you suggest a good place to start?

quozl:
Thanks for asking for directions.  I'm release manager, and oversight board chair for Sugar Labs.

My [“How to get started as a Sugar Labs developer”](http://lists.sugarlabs.org/archive/sugar-devel/2018-November/055934.html) assumes you are a self-directed developer.  You will decide for yourself what is best for you to do.  You will use Sugar, Sugarizer or Music Blocks, and choose what to maintain or improve, based on;

* your time,
* your interest,
* your equipment,
* how many other people are also interested, and;
* your previous experience with the languages and tools.

You are always the best judge of these things.  If you fail in one of these areas, you can self-improve.

Lionel (@ llaske) can tell you what needs doing in Sugarizer.  Walter (@ walterbender) can tell you what needs doing in Music Blocks.  For Sugar, what needs doing is;

* maintenance of activities, through to release and publishing,
* back-porting Sugarizer activities to Sugar,
* maintenance of the GTK+ 3 toolkit and shell (for experts).

It is tempting to do something small like the web site.  Please don't.

New developers should avoid changes to our infrastructure, such as web sites and documentation, until they have demonstrated their contribution to and awareness of the source code of Sugar, Sugarizer or Music Blocks.

This is also a cultural difference.  Sugar Labs is closer to the innovation culture of Silicon Valley than to a university.  We don't direct ourselves much at all.

For personal development in the open source innovation culture, please also read [How To Become A Hacker by Eric S. Raymond](http://www.catb.org/esr/faqs/hacker-howto.html).

savezsiddiqui:
i want to do this!

therajaryan:
I'd like to do this!

quozl:
I've said my piece already for this issue.

[...]

PR #304, opened 2019-10-09
Back To Top button

therajaryan:
Fixed issue #289 .
Added a back to the top button using JQuery and CSS on the Index page.


quozl:
No thanks.  It is not necessary.

PR #307, opened 2019-10-19
added inline css to solve the about page-imporvement issue

Sanatan-Shrivastava:
-  Followed the rules of contributions

- Added inline CSS to <p> tags separately to align the <hr> tags

- Didn't add media queries for smaller devices, Looks fine on mobile screens.


quozl:
Thanks, but I don't think this is important. The site is dead, as it has too little recent content.

Sanatan-Shrivastava:
Alright, thanks!

PR #309, opened 2019-12-29
Added spacing after list of elements 

Swayamsvk:


quozl:
Thanks, but this is not important to fix.  It is only an appearance problem.  More important is to verify and update the content on the web site, and remove content that is not maintained.  See also https://github.com/sugarlabs/www-sugarlabs/issues/305 and other closed issues for related discussion.

The content you are changing in this pull request was added during GSoC 2017 by @geekrypter but hasn't been reviewed by the oversight board.  I thought it was placeholder text, as it doesn't reflect our organisation at all.  Aspirational noise.  I've found the origin; it was copied from an old and unmaintained Wiki page; https://wiki.sugarlabs.org/go/Sugar_Labs

Issue #316, opened 2020-07-02
Port site to Bootstrap 4 from 3


samswag:
Some of the UI elements are broken and to better fix it, it will be great to migrate the bootstrap from the old version to the newest. 

Since the lastest bootstrap is a major rewrite from 3, there might be some visual changes that need to be further modified. 
See [Migration](https://getbootstrap.com/docs/4.0/migration/) on the official bootstrap website.

himanshujaidka:
So how we could migrate?

samswag:
As explained here via the official bootstrap website,  https://getbootstrap.com/docs/4.0/migration/ since the 4+ is a significant rewrite from 3. 

One major issue I have seen is the effect it has on the navigation bar because some elements changed. And to make it much easier and faster, I recommend using [BootstrapCND](https://getbootstrap.com/docs/4.5/getting-started/download/#bootstrapcdn) instead of downloading it.

himanshujaidka:
ok I got it , I am working on this

AdithyaKrishnaK:
I would like to work on this. But I was wondering how to remove the existing bootstrap files and also I am having trouble connecting to #sugar IRC channel(says Cannot send to nick/channel).

srevinsaju:
@AdithyaKrishnaK please follow the _registration_ on Freenode. :)

srevinsaju:
[Bootstrap v5.0](https://getbootstrap.com/docs/5.0/getting-started/introduction/) has been released. So probably, we would now aim to migrate bootstrap3 website to bootstrap5 now

S-kus:
I would like to work on this. @srevinsaju Are you planning to migrate it?

srevinsaju:
> I would like to work on this
Sure, you don't need to ask.

 >  @srevinsaju Are you planning to migrate it?

No. My comment was to mention that a migration to Bootstrap 5 would be better, since BS5 was released then.

S-kus:
okay, @srevinsaju. I have started working on it.

S-kus:
@srevinsaju I am planning to use the documentation of bootstrap firstly migration from [v3 to v4](https://getbootstrap.com/docs/4.0/migration/) then [v4 to v5.](https://getbootstrap.com/docs/5.0/migration/).
This will change a few files. Could you confirm this process is right? and should I commit after migrating all the things to v5 or should I do a file vise commit?

quozl:
[...]

PR #317, opened 2020-07-02
New Colours for Activities Icons.

Thusal06:
Look at **the colors of activities icons** designed by me.

### DIFF
--- a/feature_request.md
+++ b/feature_request.md
@@ -1,0 +1,13 @@
+Hi, Look at **the colors of activities icons** designed by me.
+
+These are my pictures,
+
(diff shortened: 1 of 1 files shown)


quozl:
These are not the Sugar colours.

The images do not render when the file is viewed.

samswag:
I agree with @quozl.  Instead, if you fancy, you might want to look into #316.

Thusal06:
No, the pictures are rendering. Just cut and paste the links.

Thusal06:
> These are not the Sugar colours.
> 
> The images do not render when the file is viewed.

I didn't know, you are supposed to use Sugar colors. Sorry!

quozl:
I'm surprised you didn't notice.  Which version of Sugar have you been using?  See also discussion in https://github.com/sugarlabs/www-sugarlabs/issues/174

PR #322, opened 2020-07-05
 issue#321 Social accounts open in new tab

himanshujaidka:
Now, it opens in a new tab

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -95,10 +95,7 @@
                                 </div>
 -->
                                 </br>
-                                <div class="row">
-                                    <div id="contributeHeader" class="col-md-12">
-                                        <h4 class='align-center-md'><a href="https://github.com/sugarlabs/www-sugarlabs" target="_blank"><b><span class="ion-social-github" style="font-size:17px;color: #0099FF; margin-right: 12px;"></span><span style="font-size:17px;color: #0099FF;">Contribute to website</span></b></a></h4>
-                                    </div>
+                                
                                 </div>
                             </div>
                         </div>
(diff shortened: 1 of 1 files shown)


quozl:
Why would anyone want that?  Forces the user to close one more tab.

himanshujaidka:
Because it will easy for users who wants to compare one tab with another and many users will forget to close first tab that is sugarlabs tab and this website will remain open

quozl:
Then they can use open in new tab feature.

himanshujaidka:
if they are using a mobile phone then

himanshujaidka:
In my point of view, it is more convenient for the user when the link automatically opens in a new tab :)

quozl:
This in conjunction with other recent issues and pull requests looks like a denial of service attack on our organisation.  Could you please explain why you have become interested in our web site?  Has someone asked you to be involved?  Is there a bounty?  Are we on some list somewhere?  I'd like to get us removed from any such list.

himanshujaidka:
No! nothing like that :) Actually I want to take part in GSoC 2021 so for that I am preparing and I am very passionate about it. Would you mind if you will guide me and I also want to improve your website because I love to design websites :)

quozl:
Thanks!  Looking at your profile, we could really do with help on Sugar, Sugarizer and Music Blocks.

himanshujaidka:
ok

olawanlejoel:
But I believe opening social accounts on a new tab is very useful as it helps uses still maintain what they were previously going through. I know users can easily take the 'open on a new tab' route but would be cool if it's integrated. 

The contributor can also add `rel="noopener"` with the `target="_blank"`

samswag:
@olawanlejoel I agree too, but at the moment we are focusing on fixing some broken UI elements on the site. See #316

himanshujaidka:
I think it is also a part of UI @samswag

himanshujaidka:
ok @olawanlejoel I got it

PR #326, opened 2020-08-12
Fixed some major typos

olawanlejoel:
I was reading through and I discovered few typos ( some were explicitly obvious while the others weren't).


samswag:
Thanks for the PR

PR #327, opened 2020-10-01
Improved grammar

Neeraj-Kumar-Singh:


quozl:
Thanks.  Reviewed.  I don't think any of these are an improvement.  The apostrophe after _contributors_ is correct.  The additional _from_ is not needed and makes reading odd.

PR #328, opened 2020-10-01
Added target propery and rel="noopener"

olawanlejoel:
I added the target _blank property and rel="noopener" to all links which leads outside this website. I was going through the website and decided to click a link and wanted to return to the website just to discover everything that happened was on the same tab.


quozl:
Thanks.  This change is not needed.  Should you find yourself in a similar situation, use the [Back button](https://en.wikipedia.org/wiki/Web_browser#Features) of your web browser.

PR #337, opened 2021-07-23
feat: add 501(c)(3) and tax exempt status on the website

srevinsaju:


chimosky:
Reviewed, thanks. Could you also edit the `About us` section and add this too while removing the reference to the SFC?

srevinsaju:
Updated ✨

quozl:
Yay.  (of approval type).

about-us.html
```diff
@@ -18,7 +18,7 @@
...
-                            <p class="customParagraphStyle">The Sugar development platform is available under the free/libre software GNU General Public License (GPL) to anyone who wants to extend it. “Sugar Labs” is a member project of the <a href="https://sfconservancy.org/" class="hrefCustomColor">Software Freedom Conservancy</a> (a non-profit foundation to produce and distribute and support the use of free software) and serves as a support base and gathering place for the community of educators and software developers who want to extend the platform and who have been creating Sugar-compatible applications.</p>
+                            <p class="customParagraphStyle">The Sugar development platform is available under the free/libre software GNU General Public License (GPL) to anyone who wants to extend it. “Sugar Labs” is a registered USA 501(c)(3) tax-exempt, not-for-profit organization, who serves as a support base and gathering place for the community of educators and software developers who want to extend the platform and who have been creating Sugar-compatible applications.</p>
```
chimosky, line 21: Reviewed, I think `and serves` is better than `who serves`. @aperezbios what do you think?

aperezbios: I think "which serves" is better. I'll change it in a subsequent commit.

Issue #366, opened 2022-02-03
Slow load due to CDN


quozl:
Page load using Firefox in developer tools console mode shows some resources are not used from our server; this creates external dependencies and additional domain name resolution time.

* [ ] https://code.jquery.com/jquery-2.2.4.min.js (https://github.com/sugarlabs/www-sugarlabs/issues/349),
* [ ] https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0/js/bootstrap.min.js
* [ ] https://cdnjs.cloudflare.com/ajax/libs/waypoints/2.0.3/waypoints.min.js

ishubham21:
I'm on it.

ishubham21:
How about I include minified versions of the files like `bootstrap.js` and `bootstrap.css` alongside addressing the issue?

PR #368, opened 2022-02-04
master: Issue #366 update URLs to load from local assets instead of CDN

ishubham21:
I found out that the files that were mentioned in the Issue #366 were already present as local assets. Therefore, linked to them. 
- replaced CDNs by local files
- update "https://www.sugarlabs.org/assets/favicon_06.png" to "{{ site.baseurl }}assets/favicon_06.png"
- Add bootstrap.js.map


quozl:
In future please;
* follow our guidance for [writing commit messages](https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#making-commits),
* add a "fixes" link to your opening pull request comment so that the issue is automatically closed.

Issue #349, opened 2022-01-14
Upgrade jquery


quozl:
* [x] jquery 2.2.4 is in `js/jquery.js`, but latest is 3.6.0,
* [x] _layouts/base.html uses a CDN instead of local copy, we should use our local copy.

Priyanshi-Raj:
Hi, can I work on this issue?

chimosky:
> Hi, can I work on this issue?

Hey @Priyanshi-Raj yes you can work on the issue, you should also take a look at our [contributing guide](https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md) and open a PR when you have a fix.

nikkuAg:
Hi, can I work on this issue if no one has been assigned.

chimosky:
> Hi, can I work on this issue if no one has been assigned.

Hey @nikkuAg yes you can work on the issue, see my response above too.

nikkuAg:
Hey @chimosky,

Can you please review the PR which is regarding updating jquery to the newest version.
PR: #369

quozl:
You forgot a "fixes #349".  :grin:

PR #369, opened 2022-02-05
Updating the version of jquery used

nikkuAg:
Issue: #349,

I have updated the jquery to the newest version available by changing the file `js/jquery.js`


quozl:
Thanks.

* [ ] please remove the commits unrelated to the change; it looks like you may not have rebased your branch after master gained more commits,

* [ ] please change the commit message to show how jquery.js was changed; what method you used to change it, so that someone who doesn't know how to do it can check what you did.

quozl:
Okay, but next time please keep working in the same pull request by changing the head of your branch instead of starting a new pull request.

PR #373, opened 2022-03-23
Updated the jquery to newst version3.6.0 

nikkuAg:
I have replaced the older code from the `js/jquery.js` file to to newest code from js website which is of latest version 3.6.0.

Issue: #349,


quozl:
Thanks.

I've verified it is the same as https://code.jquery.com/jquery-3.6.0.min.js

The commit message doesn't meet our guidelines here https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#making-commits

so I shall rewrite it.

PR #381, opened 2022-09-06
Remove blog

quozl:
- the planet.sugarlabs.org no longer has up to date content,

- the web site posts are no longer rendered.

Signed-off-by: James Cameron <quozl@laptop.org>


chimosky:
Makes sense, I think these should be moved to an archive page on the website and not just removed.

quozl:
Why?  They will continue to exist in git should we need them, and at the moment they give a false impression of Sugar Labs.

chimosky:
> Why? They will continue to exist in git should we need them, and at the moment they give a false impression of Sugar Labs.

They'll continue to exist in git for sure but not everyone would use git to get that information, it does give a false impression of Sugar Labs hence I think it should be archived.

quozl:
What do you mean by "moved to an archive" and "archived?"

chimosky:
> What do you mean by "moved to an archive" and "archived?"

By both terms I mean move the content to an archive page on the website.

quozl:
I guess I still don't understand.  I think the content is already on a page that people could assume is an archive, based on the dates.  How would you make that different?

chimosky:
> I guess I still don't understand. I think the content is already on a page that people could assume is an archive, based on the dates. How would you make that different?

Looked at the website and it seems it was already in an archive page and that was my suggestion seeing as keeping it under such page would clearly make it obvious that it's an archive and we have nothing to lose by leaving the archives there if you ask me.

I'm fine with the change if other people are.

quozl:
I don't think we'll get anyone else to answer, so I'll close the pull request.

PR #383, opened 2022-09-06
Drop socialhelp

quozl:
It was decommissioned many years ago.


booting-soas.html
```diff
@@ -190,9 +190,7 @@ <h3 class="text-center">STEP 6: Boot Sugar <nobold>(Traditional Boot)</nobold></
                                 <div>
                                     <img id="imgrflt" src="{{ site.baseurl }}/assets/browse-palette-ss.png" alt="Step 7 screenshot" class="customImageStyle">
                                     <h3 class="text-center">STEP 7: Enjoy Sugar on a Stick</h3>
-                                    <p>Once you selected to boot from your Sugar USB, you should be directed into the Sugar OS interface.
-                                    <br>
-                                    If you have problems during the boot process, you can check the Sugar Labs <a href="https://socialhelp.sugarlabs.org/" target="_blank" class="hrefCustomColor">Social Help</a>.</p>
```
chimosky, line 195: I think we should refer them to our matrix channel.

quozl: That would be nice.  Please make a pull request.  Meanwhile, we should remove Social Help.

chimosky: Done in 314c283.

PR #384, opened 2022-09-06
Remove search

quozl:
The search box regularly does not render in the correct position; at the moment, it renders just above and to the right of the Sugar Labs logo.  The search function does nothing particularly useful, it delegates to Google Search.  It would be best to leave searching to the better user experience of search engines and browsers.

- remove the search box,

- remove the search feature,

- remove the supporting styles and documentation.

Signed-off-by: James Cameron <quozl@laptop.org>

Fixes https://github.com/sugarlabs/www-sugarlabs/issues/380


walterbender:
Makes sense to me. And the changes look good. We should have another set ot eyes do a quick review as well.

quozl:
Thanks.  Oh, also, I did test on Ubuntu 22.04 per README.md, and it seemed to work.

chimosky:
Reviewed and tested, looks good. Thanks!

PR #385, opened 2022-09-18
Social accounts open in new tab

iamravi1:
Social accounts open in new tab


quozl:
Thanks.

Please add https://github.com/sugarlabs/www-sugarlabs/pull/386/commits/281d1b4a35ff04aecdabe8c394071999e1afb8e4 to this branch and we can close the other pull request.

I see what you've done, and how you've done it, but I don't understand why you've made the change.  How do you justify the change?

PR #387, opened 2022-09-19
Remove sidebar social media icons

quozl:
@Jyoti28Dhikale please review.

Part of a fix to https://github.com/sugarlabs/www-sugarlabs/issues/345

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -59,7 +59,7 @@
                     <div class="col-md-6">
                         <div class="row justify-content-center">
                             <div class="row d-flex justify-content-center">
-                                <div id="own-ssk" class="ssk-group d-flex">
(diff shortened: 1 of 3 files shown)


chimosky:
Not tested but I thought the aim was to improve the social links sidebar and not remove it entirely.

quozl:
The icons remain elsewhere on page, it's just the floating sidebar icons that are removed.

https://github.com/sugarlabs/www-sugarlabs/issues/345 has some of the discussion.

This pull request is proposed as a step toward the goal.  The step is an improvement because it fixes the unpleasant appearance without unduly affecting navigation or reading of the content.  Arguably, it reduces distraction too.

chimosky:
Tested, works as expected.

Issue #305, opened 2019-10-09
Use an external theme without customisation


quozl:
Much of the noise and wasted effort on our web site has arisen from using themes or stylesheets that are copied into our site and then customised extensively.  This isn't necessary.  Instead, we should use an upstream, well-known, and maintained theme, without customising it.  This may let us concentrate on content rather than form.

amaaniqbal:
+1, I mentioned about this indirectly in the past. Bootstrap v4.3 would replace many of the obsolete stylesheets we are currently having. Bootstrap is the world's most popular framework to develop responsive websites. 

It's maintained [here](https://github.com/twbs/bootstrap/issues).

quozl:
Thanks.  On the other hand, I don't want to cause more noise than signal.  In past year we have had about four content changes, out of 74 commits.  The larger problem we have is very little content being generated. :grin:

amaaniqbal:
True! Have you thought of any workflow to solve either of theme or content related problems?

By the way which content is actually being referred to here? The one on the SugarLabs website?

quozl:
> Have you thought of any workflow to solve either of theme or content related problems?

I don't understand the question.  It seems very vague.  We don't need a workflow to solve problems.

> By the way which content is actually being referred to here? The one on the Sugar Labs website?

I'm amazed at your question.  What other content could I be referring to?  Content is information about Sugar Labs, our products, our projects, our events, and our people.  Themes or stylesheets is the implementation of the arrangement of the content.

As fewer and fewer people _do anything_ at Sugar Labs, our content flow necessarily shrinks, and the content that is hosted on the site ages and reduces in relevance.  This is not a technical problem, but a social one.  It cannot be solved by technical means.

amaaniqbal:
> We don't need a workflow to solve problems.

I meant have you thought of any priorities of solving problems. In the sense, what should be targeted first and to what extent or which all tasks are required to be done ASAP, etc?

> What other content could I be referring to?

:sweat_smile: I thought you could be referring to some regular content writing for newsletters or something else apart from that on our website which is directly presented to our users. Sorry!

quozl:
> I meant have you thought of any priorities of solving problems. In the sense, what should be targeted first and to what extent or which all tasks are required to be done ASAP, etc?

No.

Thinking about in, I do notice one critical absence of content.  We're a software project, but we don't talk about what we do now.

[...]

PR #391, opened 2023-02-11
Fix the navbar padding

SrijanVerma04:
Fixes the content in the navbar with equal padding

### DIFF
--- a/css/airspace.css
+++ b/css/airspace.css
@@ -132,7 +132,7 @@ header .navbar-default .navbar-nav {
 }
 header .navbar-default .navbar-nav li a {
   color: #333333;
-  padding: 10px 26px;
+  padding: 10px 8px;
   font-size: 15px;
 }
 font header .navbar-default .navbar-nav li a:hover {


quozl:
Thanks, a nice simple change.  Though I don't know why it works.

What do you think of https://github.com/sugarlabs/www/issues/305?  First commit in this repository was a 2017 Google Summer of Code project, which took airspace.css from what is now https://github.com/ndrewtl/airspace-jekyll and made some small changes.  A diff between airspace-jekyll and that original airspace.css from 725368d3bc881bf70b9bf84cf4d5a878426f1035 shows a small degree of change.  Since then we have tended to solve layout and formatting problems by tweaking this further.  Your work on this pull request is another in that long sequence.  Can you think of a way to bring our site back into a situation where we don't have to keep changing the upstream source code in airspace.css?  Does it actually look okay with unmodified airspace.css and the other files?  Should we be using a completely different Jekyll theme now?

SrijanVerma04:
Since I am very new to open source and contributing to the project and just fix the first issue which i think by changing some of the padding can solve the problem. Since appearance is also plays the big role for the website. And for using a different jekyll theme , i think i need to spend some more time before giving any of my opinion.
So with the guidance it will be helpful what to be done for the betterment of this project!!

quozl:
Thanks.  I don't mean a different theme, but the same one ... just without our customisation or derivation.

SrijanVerma04:
okay but for now i don't have any idea How this can be done but i will look into this.

PasunuriSrinidhi:
> Since I am very new to open source and contributing to the project and just fix the first issue which i think by changing some of the padding can solve the problem. Since appearance is also plays the big role for the website. And for using a different jekyll theme , i think i need to spend some more time before giving any of my opinion. So with the guidance it will be helpful what to be done for the betterment of this project!!

Using different Jekyll themes can provide many advantages, including easy customization, pre-built layouts, responsive design, SEO optimization, integration with Jekyll plugins, and community support.

Issue #389, opened 2023-01-26
[Enhancement] : The Contents In the Navbar does not have equal padding.


CodeWizarz:
![Screenshot from 2023-01-26 13-24-59](https://user-images.githubusercontent.com/122372759/214784196-9277fa55-ec6d-4329-8f54-0cc4f69238b7.png)

The contents in a navbar or header may be designed with a specific hierarchy in mind, with certain elements intended to stand out more than others. In order to achieve this, the padding and spacing between elements may be irregular. This allows for a more dynamic and visually interesting layout, as well as making it easier for users to quickly identify important elements within the navigation. Additionally, sometimes irregular padding and spacing is used to indicate different sections or types of content. Ultimately, the goal is to create an intuitive and visually appealing navigation experience for the user.
@quozl Can you please look into this, thanks in advance.

quozl:
Thanks, but I disagree.  Our critical issue with the web site is the out of date content, and that activity by Sugar Labs developers is not being captured as new content.  Given the critical issue, the web site style and layout really don't matter.  We should be using a theme that is used by other web sites, in order to maximise uptake of our content, minimise effort to maintain, and properly direct volunteers to our core products.  Our web site is not our core product.  We should not be modifying the theme, style, or layout.

DevBobby:
@quozl I understand your concerns about the outdated content and the lack of engagement from Sugar Labs developers. However, I believe that a visually appealing and user-friendly website will also play a crucial role in attracting and retaining visitors, as well as promoting our core products effectively. Perhaps we can consider a balance between addressing the content issue and updating the theme to align with industry standards.

The-Priya-Saw:
can I work on this issue?

quozl:
No, content has priority.  Why work on an outdated web site?   I've just had a look at nav.html and most of the links are out of date.  The content must be fixed first.  If you're looking for something to do, https://github.com/sugarlabs/www-sugarlabs/pull/387 needs review.

DevBobby:
@quozl I don't intend to do this but how about we use docsaurus to create a new site other than using those bootstrap and html5. I can make a dummy webpage with the contents of sugarlabs using docsaurus I'm good with js. If it looks promising we can use the webpage we(don't mind including members) created and migrate to the new one. If it's ok with you.

quozl:
Thanks, but I don't see how that would help.  Have you reviewed https://github.com/sugarlabs/www-sugarlabs/pull/387?

[...]

PR #394, opened 2023-02-22
solving issue #389

classr-22:
Added equal padding


quozl:
That file is from the Bootstrap project.  We should not try to customise our copy of it, as that will cause problems when upgrading Bootstrap.

PR #407, opened 2024-01-13
Some tips based on one person's experience

pikurasa:


chimosky:
The problem with 5afb031 is that if any of those steps change in the future, we'll have to update it again, and it's easier if the instructions for troubleshooting or fixing issues with gem aren't handled by us.

quozl:
While true, it does work, and I've just tested something like it, so I'm happy to merge.  I'd prefer we switch to static content, as I'm not seeing the benefit of using Jekyll[^1], but until that happens we have to keep dealing with it.

[^1]: Jekyll converts Markdown and templates to a static site for us to deploy.  We rarely change the input, the Markdown and templates, and so it is hard to justify keeping Jekyll.  The time we waste on it is not recovered.

PR #408, opened 2024-01-13
Linkify our two CTAs at the bottom

pikurasa:


chimosky:
Looks good, I think the volunteer link should point to https://wiki.sugarlabs.org/go/Sugar_Labs/Getting_Involved instead of pointing to the donate page.

pikurasa:
> Looks good, I think the volunteer link should point to https://wiki.sugarlabs.org/go/Sugar_Labs/Getting_Involved instead of pointing to the donate page.

That was a mistake. Sorry about that. It's should be fixed now, pointing to https://wiki.sugarlabs.org/go/Sugar_Labs/Getting_Involved

PR #410, opened 2024-01-19
Copyedits more page

pikurasa:
This has copyedits, suggestions, and comments. The comments are in the HTML itself.

This needs review before publishing.

I have even more suggestions, but I'm trying to just make incremental improvements until we have a chance to agree upon some terminology and a styleguide.


quozl:
Reviewed.  Looks great.  Comments can be removed in a later commit.

PR #417, opened 2024-01-25
Dynamic year in copyright section added 

AdityyaX:
This pull request eliminates the manual effort required to update the copyright year in the footer of our website annually.

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -77,7 +77,7 @@
                 <div class="row" >
                     <div class="col-md-12" style="margin-top:12px;">
                         <p style="font-size:13px;" class='text-center-md'>
-                        Copyright &copy; 2024 Sugar Labs&reg; available under the <a rel="license" href="http://creativecommons.org/licenses/by-sa/4.0/">Creative Commons Attribution-ShareAlike 4.0 International License(CC BY-SA)</a>.</br>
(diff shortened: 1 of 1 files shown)


quozl:
Copyright law in many jurisdictions would frown upon an automated change that makes a legal claim to have changed content yet without any verification that content has actually changed.  To do this correctly, your code will have to determine _if_ the content has been changed before changing the year.

On the other hand, if our purpose was to make a web site that showed you the current year, what you suggest would be fine.  :grin:

quozl:
Oh, and the way you have written it depends on the year being set correctly on the computer running the web browser, rather than on our server.

AdityyaX:
Is there any proposed changes, or i shall leave this pull request as it is right now?

quozl:
Thanks.  I've given my views.  I'd like to hear from other members of Sugar Labs.

chimosky:
> Copyright law in many jurisdictions would frown upon an automated change that makes a legal claim to have changed content yet without any verification that content has actually changed. To do this correctly, your code will have to determine if the content has been changed before changing the year.

I agree with Quozl on this, the website doesn't change so much and regularly so I'm a bit skeptical about this.

pikurasa:
We appreciate this, but we will not be doing dynamic copyright years.

Issue #418, opened 2024-01-26
Local labs broken links


pikurasa:
These links are currently broken:

- http://co.sugarlabs.org/go/Laboratorios_Az%C3%BAcar_Colombia
- http://cl.sugarlabs.org/go/P%C3%A1gina_Principal
- http://pe.sugarlabs.org/ir/
- http://ar.sugarlabs.org/

These are all linked from https://www.sugarlabs.org/about-us/

What should they be updated to? Or should we just remove them for now?

quozl:
I suggest removing.

chimosky:
> I suggest removing.

I agree.

pikurasa:
Removing the entire section or just the broken links? (If we remove the broken links, it leaves only one link for "Local Labs")

sanyamk0:
> Removing the entire section or just the broken links? (If we remove the broken links, it leaves only one link for "Local Labs")

Hello I am Sanyam a 3rd Year Undergraduate from IIIT Bhopal, India.
I guess it will make sense to remove only broken links and not the link that's working fine.

One more thing ig those broken links are for different countries for translation purpose and better way would be fixing deployment of those links.
(And also can't we just keep a single website for all countries and it will render language according to country site is operated.)

Please correct me if i'm wrong anywhere, I'm new to Open Source and trying to give a valuable contribution to organizations.

sanyamk0:
Also if the links needs to be removed, I would love to contribute to it.
And also in integrating different language to this site dynamically.

chimosky:
> Removing the entire section or just the broken links? (If we remove the broken links, it leaves only one link for "Local Labs")

The link to the DC local lab isn't broken, only thing we need to figure out is if the local lab is active else I don't see a need to have it when it's not active.

Removing the broken links is a good way to go, I don't think we should remove the section as the practice was a good one and we'd like it to continue albeit none of the local labs seem to be running today.

chimosky:
>? One more thing ig those broken links are for different countries for translation purpose and better way would be fixing deployment of those links.
(And also can't we just keep a single website for all countries and it will render language according to country site is operated.)

@sanyamk0 the links won't be fixed as no one is currently maintaining wherever they led to, and the links contained varying info so they can't be kept in a single page and rendered by language.

> Also if the links needs to be removed, I would love to contribute to it.
And also in integrating different language to this site dynamically.

We're leaning towards removing and that's probably what we'll do.
[...]

PR #419, opened 2024-01-28
Comment out local labs section (#418)

pikurasa:
Addresses issue #418 (in a somewhat clumsy way)


chimosky:
This is good, I'd just asked Bernie on matrix if he knows anything about those links as I can only see when one was added.

quozl:
We can always put it back after Bernie answers.

PR #431, opened 2024-05-03
It's hidden (for now), but it should make the page verified

pikurasa:
I suspect there are some updates we can run to retrieve Mastodon logo and related CSS styling, but, for now, I think merging what I have will trigger Mastodon's verification.


pikurasa:
Ok, I also figured out how to make it visible on the "Contact Us" page, which is more straightforward (no special CSS sauce).

pikurasa:
It looks like this:

Screenshot from 2024-05-02 20-28-22

pikurasa:
The footer now looks like this in the latest commit:

Screenshot from 2024-05-04 21-03-57

...which is now a placeholder until we figure out how to update the repositories of social icons.

walterbender:
I think we just need to update this file: https://github.com/sugarlabs/www/blob/master/css/ionicons.min.css

walterbender:
And we prob. want to ^twitter^x
There is a mastodon logo in the ion-icon set we can use.

pikurasa:
> I think we just need to update this file: https://github.com/sugarlabs/www/blob/master/css/ionicons.min.css

I think there are a batch of files to update. See: https://github.com/sugarlabs/www/commit/932dbab16cb5194ca033edd46448445f908c5526

I looked yesterday, but wasn't sure how to update it.

I /think/ the related code is here: https://github.com/ionic-team/ionicons

pikurasa:
This is solved now by https://github.com/sugarlabs/www/pull/437

PR #432, opened 2024-05-08
Add a draft of an announcement

pikurasa:
You can test the draft by running `bundle exec jekyll serve --drafts`

Once we get it the way we want it, we copy it over with the YYYY-MM-DD-name.md naming convention; and, then, it should automagically create the html.

### DIFF
--- a/_drafts/Sugar-Labs-announces-nonprofit-status-new-executive-director.md
+++ b/_drafts/Sugar-Labs-announces-nonprofit-status-new-executive-director.md
@@ -1,0 +1,71 @@
+---
+layout: post
+title: "Sugar Labs announces nonprofit status, new executive director"
+---
+
+[PDF](/press/Sugar-Labs-nonprofit-announcement.pdf)
+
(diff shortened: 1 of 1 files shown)


pikurasa:
And it looks like this:

Screenshot from 2024-05-08 14-29-03
Screenshot from 2024-05-08 14-28-31

quozl:
So it didn't appear on the web site.  I've run out of time for the moment, but next step is to check the automatic deployment process, or rename the file to match the convention in the hope that the date in the file name may be important.

pikurasa:
> So it didn't appear on the web site. I've run out of time for the moment, but next step is to check the automatic deployment process. Given that you got it working locally, it should be fine.

I wouldn't expect it to appear yet.

It needs to be moved from _drafts folder to _posts folder.

The workflow should be: review it "as draft", approve it, then we deploy it "as post".

pikurasa:
This is a PR for the last step of deploy as a post: https://github.com/sugarlabs/www/pull/433

PR #434, opened 2024-05-31
Update contact

pikurasa:
This updates the contact info page.

The formatting for the address is taken from https://scanmailboxes.com/support/docs/some-of-the-sender-or-the-vendor-will-not-recognize-my-new-scan-mailboxes-address-due-to-their-automatic-verification-system-what-should-i-do/

Our PMB at that address is: 1271


pikurasa:
Apparently, I hadn't checked out the master branch before working on this. Please see https://github.com/sugarlabs/www/pull/435 instead. (Sorry!)

PR #438, opened 2024-06-11
Upadate activities link to point to aslov4

chimosky:
Some activities that are aslov4 still point to aslo, and we don't want that, the Sugar Activity also pointed to aslo and this changes that.

@pikurasa kindly review.


pikurasa:
There are more activity links than just that one. Why not change them all? (Or am I missing something...)

chimosky:
Some of the activities haven't been uploaded to aslov4 so I didn't change them.

pikurasa:
> Some of the activities haven't been uploaded to aslov4 so I didn't change them.

Got it. Sorry, I didn't understand that.

quozl:
For your interest, Devin, one can imagine several queues of development;

* activities to be ported from GTK 2 to GTK 3, then they join aslo,
* activities to be ported from Python 2 to Python 3, then they join aslov4,
* activities to be packaged using flatpak for standalone use,
* activities to be ported to Sugarizer and JavaScript, then they join Sugarizer,
* activities to be ported from Sugarizer into Sugar Web activities and join aslov4.

That last queue has been inactive for a few years because Sugar Web hasn't been working when used inside Sugar.

PR #441, opened 2024-06-12
Update readme

soumyaranjan-panda:


soumyaranjan-panda:
Hi, @pikurasa @chimosky, can you help? I tried rebasing, but the previous commits are still here. How do I deal with this?

Issue #416, opened 2024-01-25
Issue: Implement Dynamic Year in Footer


AdityyaX:
Currently, the copyright year in the footer of our website requires manual updates every new year. To enhance user experience and alleviate the need for annual adjustments, we propose implementing a dynamic year feature.

quozl:
Please don't open an issue if you are planning to open a pull request. Just open the pull request instead. Issues are best used for things you are not going to fix, and hope someone else will fix them.

You also forgot to link your pull request to the issue.  I shall do that now; https://github.com/sugarlabs/www/pull/417

quozl:
Note the feature must not use the clock of the browser to determine the year; it must be set by the files in the repository.

pikurasa:
We do not plan to implement this.

PR #447, opened 2024-08-01
Dynamically update copyright year based on latest commit date

Keshav-girase:
I’m contributing to the project for the first time and have added a feature to dynamically update the footer copyright year based on the latest GitHub commit date.

1. Code: Fetches the latest commit date from the GitHub API and updates the footer year.
2. Verification: Clone the repo, check the footer in a browser, and verify the year updates correctly.

Please review the changes and guide me if there are any adjustments needed before merging.

Thanks!

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -82,7 +82,7 @@
                 <div class="row" >
                     <div class="col-md-12" style="margin-top:12px;">
                         <p style="font-size:13px;" class='text-center-md'>
-                        Copyright &copy; 2024 Sugar Labs&reg; available under the <a rel="license" href="http://creativecommons.org/licenses/by-sa/4.0/">Creative Commons Attribution-ShareAlike 4.0 International License(CC BY-SA)</a>.</br>
+                        Copyright &copy; <span id="year">2024</span> Sugar Labs&reg; available under the <a rel="license" href="http://creativecommons.org/licenses/by-sa/4.0/">Creative Commons Attribution-ShareAlike 4.0 International License(CC BY-SA)</a>.</br>
                         Website base template: Designed and developed by Themefisher
                         </p>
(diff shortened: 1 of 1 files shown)


quozl:
Fascinating.  @AdityyaX, please review.  You opened issue https://github.com/sugarlabs/www/issues/416, and you contributed an early change, so you are a great candidate to do the review.  Also, as far as I know, nobody else wants the year to be dynamic.  It's your idea, and now someone else has suggested a solution.

We also have https://github.com/sugarlabs/www/pull/417 still open without progress toward agreement.

As there, I still think the year should not be dynamic, but that won't stop me reviewing.  This pull request 447 makes the web site rely on a connection to GitHub at the time of page load, but does default to 2024.

Keshav-girase:
@quozl Thanks for the feedback. I understand the concerns about relying on GitHub for the year update. The default to 2024 ensures the site remains functional. I appreciate your review and am open to any suggestions.

quozl:
Okay, can you tell me why you think it should be done at all?  There's so much else not done, like actual content.

Keshav-girase:
I appreciate your perspective. I suggested this change to ensure the footer year is always current without needing manual updates, which can sometimes be overlooked. However, I understand that focusing on actual content is a higher priority. I'm happy to focus on other tasks that align better with our current goals.

quozl:
But why do you think the footer year should be always current?  Perhaps it should be the date the page content was last changed?  We should strive to be accurate on our web site, and that includes how current the content is.  Claiming the content as more current than it is would be deceptive.  For instance, most of index.html was written before 2018, yet the footer suggests 2024.

pikurasa:
We appreciate this, but we will not be doing dynamic copyright years.

Issue #458, opened 2024-10-11
Create better layouts for some of our sub-pages


pikurasa:
Need not be "perfect", but some improvements to the following would be nice:

- Come up with recommendations for better layout for:
- /sugar-stories/
- /2014/01/22/Sugar-Labs(R)-Educational-Nonprofit-Celebrates-Two-Google-Code-In-Grand-Prize-Winners/ etc.
- /parents/ and /school-administrators/ and /students/ (which are all similar)
- /more/

ishwarthecodddr:
assign me this issue

omsuneri:
@ishwarthecodddr we dont assign issues

PR #461, opened 2024-10-12
Improve Sugar Stories Page

soumyaranjan-panda:
Fixes: #458

Here is the comparison of the updates: the first one is the after update, followed by the before update.

image


chimosky:
Haven't tested yet, did you test on various screen sizes and be sure it produced the desired results?

soumyaranjan-panda:
No, It doesn't work as intended, I'll update in my next pr.

PR #464, opened 2024-10-12
Updated Sugar Stories Layout

soumyaranjan-panda:
fixed: #458 
Tested on various screen sizes.
Here is the comparison of the updates:
image (1)

The absence of margin-top: 80px was causing issues on screens with a width of 767px or less.
image (2)


pikurasa (approved):
Looks good, and tested.

After <-- Before

Screenshot from 2024-10-12 18-39-46

Issue #466, opened 2024-10-12
Replace Sugar Stories table with HTML table


pikurasa:
On Sugar Stories page, replace /assets/data_table.png with an HTML table

omsuneri:
@pikurasa I ll make it

sarthakvyas16:
I want to work on this issue please assign me this issue.

omsuneri:
@sarthakvyas16 the issue is already solved try to move for another issue

sarthakvyas16:
Ohkay😂

PR #467, opened 2024-10-12
FIXES ISSUE #466 Replace Sugar Stories table with HTML table

omsuneri:
Resolved png to html table.
Screenshot 2024-10-13 at 4 56 52 AM


sugar-stories.html
```diff
@@ -50,9 +50,26 @@ <h3>Story 1: The connection between Sugar - Students - Teachers</h3>
...
+                      <th>Release number (date)</th>
+                      <th>Total Commits</th>
+                      <th>Youth Commits</th>
+                      <th>Release note URL</th>
+                  </tr>
+                  <tr>
+                      <td>0.102 (July 2014)</td>
+                      <td>424</td>
+                      <td>108</td>
+                      <td><a href="https://wiki.sugarlabs.org/go/0.102/Notes">https://wiki.sugarlabs.org/go/0.102/Notes</a></td>
```
pikurasa, line 64: I'm certainly not an HTML expert, but I tried adding <p> and </p> around the text in these areas, and it looks better, especially on a larger screen. Please try this.

pikurasa (changes requested):
Please try improving the HTML in the table.

A start is to add `<p></p>` to the text inside the sections of the table.

pikurasa:
After doing some research, I found that if you add `style="width: 100%; "`, then it fills the width of the screen.

`<table style="width: 100%; " class="customMargin6">`

The result looks like this:
Screenshot from 2024-10-13 15-29-06

Otherwise, it doesn't fill the screen's width. Right now, yours looks like this:

Screenshot from 2024-10-13 15-31-10

I figured this out by reading https://stackoverflow.com/questions/7138605/how-do-i-make-one-table-column-fill-all-the-spare-horizontal-space

It would also be nice if the table kept a little space around the text, whether or not the screen's width is wide or not.

omsuneri:
@pikurasa I had tried it with the width 100% but it is resulting in something more increased table than the paragraph width so it's not that much pretty and I thought with the last commit merged the paragraph width has been increased so that's why my ss is different from yours but no issues I ll update my pr according to requirements

pikurasa:
@omsuneri Is this ready for review again?

PR #473, opened 2024-10-13
Fixed #465 Story1 and Reference Button Non-Functional on Width 991px to 1214px

### LINKED ISSUE #465: Story1 and Reference Button Non-Functional on Width 991px to 1214px
### Description

I have encountered an issue where the **Story1** and **Reference** buttons are not functioning properly when the screen width is between **991px** and **1214px**.

### Steps to Reproduce

1. Open the application on a device or emulator with a screen width between **991px** and **1214px**.
2. Navigate to the `/sugar-stories` section containing the **Story1** and **Reference** buttons.
3. Attempt to click on either button.

### Expected Behavior

The buttons should be clickable and perform their intended actions.

### Actual Behavior

The buttons do not respond when clicked.

igennova:
Summary
Fixes an issue where the Story1 and Reference buttons were not functioning between screen widths of 991px and 1214px.

Problem
The buttons in the /sugar-stories section did not respond to clicks when viewed on devices with a screen width between 991px and 1214px. This impacted the user experience as the buttons were not usable in this specific range.

Fix
Adjusted the z-index and position properties in the CSS:

Added z-index: 10;
Set position: relative;
This ensures the buttons now respond properly to clicks in the specified screen width range.

Fixes #465

### DIFF
--- a/css/airspace.css
+++ b/css/airspace.css
@@ -1850,6 +1850,8 @@ px      68    160    252    344    436    */
 
 
 #exTab3 .nav-pills > li > a {
+  z-index: 10;
+  position: relative;
   border-radius: 30px 30px 30px 30px ;
  border: 1px solid #474354;
  -webkit-box-shadow: 0 3px 6px rgba(0, 0, 0, 0.16), 0 3px 6px rgba(0, 0, 0, 0.23);


pikurasa (approved):

pikurasa:
It works.

It's not an entirely robust solution, because for sizes smaller than 991 the buttons are not visible/functional at all.

That said, it's a step in the right direction.

chimosky:
> It works.
> 
> It's not an entirely robust solution, because for sizes smaller than 991 the buttons are not visible/functional at all.
> 
> That said, it's a step in the right direction.

If that's the case then an issue for that should be open seeing as the bug was introduced by this change.

Issue #477, opened 2024-10-13
Some favicons are not rendering (404 error)


pikurasa:
I refreshed sugarlabs.org and got the following error:

```
www.sugarlabs.org/assets/favicon_010.png:1 
        
        
       Failed to load resource: the server responded with a status of 404 ()
```

I think that we may have a logic issue in `_site/js/main.js`, and I'll be submitting a PR that needs review.

pikurasa:
I think, on `_site/js/main.js' that the code in the first 13 lines should be:

```
$(document).ready(function(){
    // Colour changing affect of logo
    var COLORS = [['006E00', '78E600'], ['68008C', '78E600'],
                  ['006E00', '00BEFF'], ['033CD2', '00BEFF'],
                  ['6E008C', '00BEFF'], ['A00000', 'FF7800'],
                  ['033CD2', 'FF64D2'], ['6E008C', 'FF64D2'],
                  ['6E008C', 'FF7800'], ['033CD2', 'FF7800'],
                  ['006E00', 'FFD732'], ['6E008C', 'FFD732']];
    var color = Math.floor((Math.random() * 12));
    var logoID = color + 1;
    if (color < 9) {
        logoID = "0" + logoID;
    }
```

(if color is is greater than 9, not 10, then prepend with "0")

However, it seems that that file is on the gitignore list, so making the change will add a lot of unrelated changes (at least if done from my local repo).

pikurasa:
> However, it seems that that file is on the gitignore list, so making the change will add a lot of unrelated changes (at least if done from my local repo).

I think I figured out what was going on there. I made the fix to `/js/main.js` in the PR.

PR #479, opened 2024-10-13
Prepend with zero if less than 9 not 10, since we are adding 1

pikurasa:
This is intended to solve https://github.com/sugarlabs/www/issues/477

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -8,7 +8,7 @@ $(document).ready(function(){
                   ['006E00', 'FFD732'], ['6E008C', 'FFD732']];
     var color = Math.floor((Math.random() * 12));
     var logoID = color + 1;
-    if (color < 10) {
+    if (color < 9) {
         logoID = "0" + logoID;
     }
     document.querySelector('#defaultIcon1').href = 'https://www.sugarlabs.org/assets/favicon_' + logoID + '.png';


js/main.js
```diff
@@ -8,7 +8,7 @@ $(document).ready(function(){
                   ['006E00', 'FFD732'], ['6E008C', 'FFD732']];
     var color = Math.floor((Math.random() * 12));
     var logoID = color + 1;
-    if (color < 10) {
+    if (color < 9) {
```
walterbender, line 11: < 9 is 1-8. < 10 is 1-9. I think we want the latter.

pikurasa: @walterbender 10 is what we already have; we are getting this error: https://github.com/sugarlabs/www/issues/477; so I suspected it may be due to the `color +1`

chimosky (commented):
Looks good, I think it should be a `<= 9` as that'll include the 9.

pikurasa:
I'm suspecting that the problem is, and the reason we run into https://github.com/sugarlabs/www/issues/477, is because we add one before prepend zero operation, which tests if the number for "color" is greater than 9.

`    var logoID = color + 1;`

However, I'm closing this, because I think it will probably be easier for everyone to follow the logic if we do the test for greater than 9 against "logoID", not "color", since the former is the result after adding 1. I've tested this, and it seems to work.

PR #480, opened 2024-10-14
fix: remove redundant JS loop and limit press rendering to HTML

### LINKED ISSUE #456: Beautify /press/ and figure out how to limit to show only the last 10, then put the other "archives" on another page.
Two things:

1. make the /press page look better
2. limit entries to "10 most recent" and put other pages on an archive -- if this is very difficult, let me know. It's not super important, I suppose.

haroon10725:
Fixes #456

There was an unnecessary loop in the JS file to limit the displayed press, since the HTML already has a loop for this purpose. I believe we should only use the HTML loop to limit the posts. I think this approach is much better.

@pikurasa @chimosky What do you think about it ?


chimosky:
Haven't reviewed or tested your change, but your opening comment should be part of your commit message because git stores your commit message which anyone can access, but it doesn't store your Github opening comment.

Please see [making commits](https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#making-commits).

PR #495, opened 2024-10-18
fix: added css property for better link visibility

### LINKED ISSUE #492: Make our link an easier-to-read color
The styling used for the link under "Check out the remaining hundreds of activities!" makes it hard to see the link.

Screenshot from 2024-10-17 20-35-46

It should be white and highlight only upon hover, just like the links in the footer.

We're probably just calling the wrong CSS.

haroon10725:
Fixes #492

https://github.com/user-attachments/assets/04913d32-8975-40ee-af65-fa99f3b897db


haroon10725:

PR #504, opened 2024-10-19
feat: redesign activities section for improved layout and user experience

soumyaranjan-panda:
- Redesigned the activities section for improved layout and user experience.
- Added hover animation for interactive elements.
- Tested responsiveness across different screen sizes.
- Reduced icon sizes as they don't contain detailed data, making them more appropriate.
- Fixed background color inconsistency to ensure the remaining activities section matches and feels cohesive.

image

Note: I've changed some captions; if necessary to change, please let me know.

### DIFF
--- a/index.html
+++ b/index.html
@@ -438,105 +438,90 @@ title: Sugar Labs
   <div class="container">
     <div class="row">
       <div class="section-title">
-        <h2>Already using Sugar? We have many activities for you!</h2>
-        <p>The Sugar Learning Platform is a complete environment for teaching and learning, which includes individual activities. If you're already using the Sugar Desktop Environment, then you can install from the activities below, which has links to some of our most popular activities.</p>
+        <h2>Explore Sugar Activities</h2>
+        <p>Discover a world of interactive learning with our most popular Sugar activities. Perfect for students already using the Sugar Desktop Environment.</p>
       </div>
     </div>
-    <div class="row "><!-- This is one too many for this 4x2 grid and the SVG -- and only this SVG -- renders too large
+    <div class="row activity-grid">
       <div class="col-sm-6 col-md-3">
         <div class="service-item">
-	  <img class="span1" src='{{ site.baseurl }}/assets/activity-numbers.svg' alt="Numbers activity svg">
-          <br>
+	        <img class="activity-icon" src='{{ site.baseurl }}/assets/activity-numbers.svg' alt="Numbers activity icon">
           <h4>Numbers</h4>
-          <p>Play with numbers</p>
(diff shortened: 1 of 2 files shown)


pikurasa:
It's nice. However, upon testing, width of 577-769 is a bit awkward because the sections only use up a small portion of the screen's width.

Screenshot from 2024-10-19 21-47-12

I recommend that, for anything under 770, each one of these sections use the entire width of the screen like they do at ~577 and under.

Screenshot from 2024-10-19 21-48-42

soumyaranjan-panda:
image

Gemfile.lock
pikurasa, line 1: Do we need to include the gemfile.lock file?

It's a file on the list for .gitignore, so I assume not.

index.html
```diff
@@ -438,105 +438,90 @@ <h3 class="text-center">For more details, click the link below!</h3>
...
       <div class="section-title">
-        <h2>Already using Sugar? We have many activities for you!</h2>
-        <p>The Sugar Learning Platform is a complete environment for teaching and learning, which includes individual activities. If you're already using the Sugar Desktop Environment, then you can install from the activities below, which has links to some of our most popular activities.</p>
+        <h2>Explore Sugar Activities</h2>
+        <p>Discover a world of interactive learning with our most popular Sugar activities. Perfect for students already using the Sugar Desktop Environment.</p>
       </div>
     </div>
-    <div class="row "><!-- This is one too many for this 4x2 grid and the SVG -- and only this SVG -- renders too large
+    <div class="row activity-grid">
       <div class="col-sm-6 col-md-3">
```
pikurasa, line 446: I experimented with changing these classes to be "col-md-4", like the section above.

Screenshot from 2024-10-21 09-13-56

That seems to work better for using the full width of the screen.

WDYT?

pikurasa (changes requested):
I recommend making a new PR that doesn't include the gemfile.lock file.

Gemfile.lock
soumyaranjan-panda: I'll make another pr.

index.html
soumyaranjan-panda: That looks good. I'll review and implement it in my next PR.

soumyaranjan-panda: @pikurasa How about using `col-sm-6`, `col-md-4`, and `col-lg-3`?

soumyaranjan-panda: Two columns for a large window screen will take up a lot of space, so making it 3 columns is wiser. For tablet screens, 2 columns seem perfect.

pikurasa: That makes sense, but I'm pointing out something else: When it moves to one column, it doesn't use the full width.

Notice, in my screenshot above, the difference between "numbers" section and "finance".

I think, when it moves to one column, it should always fill the full width.

soumyaranjan-panda: Understood

PR #505, opened 2024-10-19
feat: redesign donation section

soumyaranjan-panda:
**Description:**
- Added background for better visual appeal.
- Fixed typos: "reoccuring" → "recurring" and "tax-deductable" → "tax-deductible."
- Combined link and button redirecting to the same page into one link.
- Replaced image button with an actual `<button>` tag. 

Please review and provide feedback if needed. Here's before after update:
image (1)

### DIFF
--- a/css/airspace.css
+++ b/css/airspace.css
@@ -247,7 +247,7 @@ font header .navbar-default .navbar-nav li a:hover {
 }
 
 #donation{
-  background-color: #4a4063;
+  background-color: #474354;
   color: white;
   padding: 60px 0;
   margin: 2rem 1.5rem;
--- a/css/airspace.css
+++ b/css/airspace.css
@@ -245,6 +245,42 @@ font header .navbar-default .navbar-nav li a:hover {
   font-size: 14px;
   line-height: 1.8;
 }
+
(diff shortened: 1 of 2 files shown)


pikurasa:
@soumyaranjan-panda I merged this in particular in order to patch the typos (thank you).

However, we should keep the indication that you can donate by credit card as well.

What do you think the best way is to do this?

We could:
- Just say "Donate"
- Say "Donate with PayPal or credit card"
- Put an image of the types of credit cards you can use underneath

WDYT?

Also, I wonder if there is a way to make the "Learn more" a little easier to read.

soumyaranjan-panda:
I initially thought naming the button 'Donation with PayPal' was appropriate since it was redirecting to paypal page. However, I agree with your suggestion to update the button caption to just 'Donate' and to add an image of the accepted card types underneath.
For the "Learn more" link, I can make it a lighter shade of blue and underline it to improve readability. Let me know if there’s anything else you’d like to adjust!

pikurasa:
> Let me know if there’s anything else you’d like to adjust!

Those updates sound good to me.

PR #516, opened 2024-10-23
fix: replace ion icons with SVGs and remove script dependencies

haroon10725:


pikurasa:
Can you please put it inside a new folder inside assets? Like assets/ion-icons ?? Or put it inside the fonts folder.

Let's keep assets to things that are SL-specific.

PR #522, opened 2024-10-27
Adding new section named reference video in MB page

omsuneri:
**Allows user to play and view videos on te MB page directly
**Also added a link with MB youtube Playlist.
After :
image


pikurasa:
Can you please move your custom CSS code to custom.css, so that we may better track our custom code?

If there is anything that makes that difficult, please let me know.

omsuneri:
@pikurasa yaa I will do it in an hour

PR #530, opened 2024-10-30
Adding new section named reference video in MB page

omsuneri:
**Allows user to play and view videos on te MB page directly
**Also added a link with MB youtube Playlist.
After :
image


omsuneri:
@pikurasa new PR is created for refrence video section

PR #531, opened 2024-11-01
fix: remove extra dropdown icon

haroon10725:
Before:
Before

After:
After

### DIFF
--- a/css/airspace.css
+++ b/css/airspace.css
@@ -2430,20 +2430,6 @@ h1:before {
     position: relative;
 }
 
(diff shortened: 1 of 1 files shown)


haroon10725:

pikurasa:
> @pikurasa Can you please review it ?

Is the code your deleting here custom code, or was it from upstream?

Or do we have an easy way of knowing?

haroon10725:
@pikurasa I think it is a custom code.

Issue #521, opened 2024-10-26
Carousels are broken


pikurasa:
It seems that all our carousels are currently broken, which effects things like our testimonial slider (on the front page).

As for as I can tell:
* It worked on dfaa771173282226556e1bb99ab8e0080d51df10
* It stops working on https://github.com/sugarlabs/www/commit/7088c6ddb1b9159eaebd8c8b4138dd83962e55f0

It's more important to have the navbar working than the slider, so if there isn't a quick fix we may opt for a different method of displaying the information.

Here's a screenshot of an inspection of the testimonial slider on the front page:
![Screenshot from 2024-10-26 11-56-05](https://github.com/user-attachments/assets/0225ff9b-9399-4e9e-8b8f-180cbb4cd130)

avanshh99:
Hello there! Please assign me this issue so that i could work on that

pikurasa:
> Hello there! Please assign me this issue so that i could work on that

We don't assign issues, but we do review PRs. Also, we recommend you join us in https://matrix.to/#/#sugarlabs-web:matrix.org, where we are coordinating efforts.

Nama21yo:
@pikurasa Please review my PR #540, which adds a Bootstrap carousel with custom CSS styling. Is there anything that should be improved?

therealharshit:
@pikurasa i have fixed it and raised a PR #541 please review.

PR #540, opened 2024-11-09
Updated the Carousel 

Nama21yo:
https://github.com/user-attachments/assets/5022d5fa-eece-4d52-94a2-8fe2bd7e18ef

Ref #521 I have added a Bootstrap carousel to the Testimonials section, with custom CSS for improved styling and responsiveness. The navigation controls and indicators are styled for better usability, and the overall design complements the existing layout without disrupting the legacy theme.

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -1,75 +1,159 @@
 <!DOCTYPE html>
 <html class="no-js" lang="en">
-    <head> 
-        <meta charset="utf-8">
-        <meta name="viewport" content="width=device-width, initial-scale=1">
-        <meta http-equiv="X-UA-Compatible" content="IE=edge">
- 		<link id="defaultIcon1" rel="shortcut icon" href="{{ site.baseurl }}assets/favicon_06.png" />
-        <title>{{ page.title }}</title>
-        <link href="{{ site.baseurl }}/css/googleFonts.css" rel="stylesheet" type="text/css" >
-        <meta name="robots" content="index,follow,NOODP" />
-        <meta name="description" Content="At Sugar Labs, we make FOSS tools that learners use to explore, discover, create, and reflect. We encourage our users to take responsibility for their learning." />
-        <meta name="keywords" Content="Sugar Labs, Sugar, Sugarlabs, Education, Opensource, OpenSource, Sugar activities, activities, Sugar OS, Sugarizer, Sugar on Stick, SOAS, FOSS, Free software, software libre, Free/Libre Software, Constructionism, GSOC, GCI, Outreachy, Sugar Software, Learning" />
-        <!-- Facebook card -->
-        <meta property="fb:admins" content="552694361,100003527081308,100003017988002,100007693992282" />
-        <meta property="fb:page_id" content="187845102582" />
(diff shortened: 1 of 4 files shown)


quozl:
Why change airspace.css?

quozl:
There's too much irrelevant change, I can't see what you've really changed, and your commit messages don't explain.

Nama21yo:
> Why change airspace.css?

I updated airspace.css to add some cool effects to the slider, like smooth transitions and hover animations, to make it look better and feel more interactive.

Nama21yo:
> There's too much irrelevant change, I can't see what you've really changed, and your commit messages don't explain.

Okay other than the HTML and CSS I have added a cdn of js files to ensure the carousel functions properly. 
```
<script src="https://code.jquery.com/jquery-3.5.1.slim.min.js"></script>
<script src="https://cdn.jsdelivr.net/npm/@popperjs/core@2.5.2/dist/umd/popper.min.js"></script>
```

quozl:
Thanks.

We'd like not to update files that come from themes or other software projects, so your change to airspace.css while in keeping with some of the earlier unfortunate work by others isn't helping toward minimising change against upstream themes.  Can you place your changes in a more appropriate place?

You've added CDN of JavaScript.  We've preferred to deliver all CSS from our own server, which has a type of CDN already.  If you've found the carousel functions improperly using our own server, we want to fix that rather than introduce more load-time dependency on an external organisation.

Please remove your changes to indentation so that the view of the pull request by files changed shows what you intend without extra things.

pikurasa:
@Nama21yo What you created visually looks nice.

What @quozl is saying about calling the JS is correct; we need to avoid that.

As for custom CSS, please put that into the custom.css file.

Nama21yo:
> @Nama21yo What you created visually looks nice.
> 
> What @quozl is saying about calling the JS is correct; we need to avoid that.
> 
> As for custom CSS, please put that into the custom.css file.

Okay the problem was with my vs code extension called prettier it formats the code and gives spacing that why it looks formatted too much and yes I shan't integrated CDN of js for carousel since it exists sorry I will correct it in my next PR on another issue.

quozl:
Okay, that's expected.  Turn off reformatting extensions for functional pull requests.  Turn on reformatting extensions for new files only, or where a pull request is _only_ for reformatting.

Otherwise review is very difficult.

therealharshit:
This issue has been fixed in bfcb246db65654d3ea822067730d39e317ecfdc0

quozl:
That's great, but please submit the changes in  bfcb246 in the form of a pull request.

I see this pull request doesn't have the change in that form, and was based on your _master_ branch.  We recommend branching before making a pull request, see https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#branching

Nama21yo:
> Oh, I see, bfcb246 is also on master branch now.  Thanks @therealharshit.  @Nama21yo do you want this pull request to proceed?
No , @quozl   
@therealharshit has resloved the issue with optimized code I will proceed on another issue.

quozl:
Thanks!  Closing.

PR #541, opened 2024-11-09
fixes #521 Carousels

### LINKED ISSUE #521: Carousels are broken

therealharshit:
fixes #521 
@pikurasa please review.

https://github.com/user-attachments/assets/2b285665-1a8a-4638-95be-24b4c6c29903

### DIFF
--- a/_layouts/base.html
+++ b/_layouts/base.html
@@ -66,6 +66,13 @@
                 $(".loader").fadeOut("slow");
             })
         </script>
+        <script>
+        $(document).ready(function(){
+            $(".owl-carousel").owlCarousel({
+           paginationSpeed : 500, singleItem:true, autoPlay: 3000,
+            });
+         });
+        </script>
     </head>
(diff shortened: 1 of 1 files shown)


pikurasa:
It fixes the problem in what seems to be the simplest manner.

therealharshit:
> It fixes the problem in what seems to be the simplest manner.

Thanks

PR #546, opened 2024-11-10
add documentation for how to create a FAQ

haroon10725:
@pikurasa Can you please review it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,3 +320,25 @@ In this section, we will guide you through the steps to add new icons. By follow
 8. **Updating Ion Icons**
    * Install the latest version of Ion Icons by running the following command:
    `npm i ionicons@7`
+
+### How to create a FAQ?
+In this section, we will guide you through the steps for creating a new FAQ.
+
+1. **Navigate to the _faqs directory:**
+   * In the site's main directory, open the _faqs folder. This is where all your FAQ posts are stored.
+
+2. **Create a New Markdown File:**
+   * Create a new file in the _faqs folder.
+
+3. **Add Front Matter:**
+   * At the top of your new file, add front matter between `---` lines. This metadata tells Jekyll how to handle the post. Here’s a basic template:
+
+   ```
+   ---
+   question: "The question you want to add"
+   answer: "The answer to that question"
+   ---
+   ```
+
+4. **Save and Preview:**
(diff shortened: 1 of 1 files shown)


therealharshit:
@haroon10725 looks good, thanks.

README.md
```diff
@@ -319,4 +319,26 @@ In this section, we will guide you through the steps to add new icons. By follow
...
+   `npm i ionicons@7`
+
+### How to create a FAQ?
+In this section, we will guide you through the steps for creating a new FAQ.
+
+1. **Navigate to the _faqs directory:**
+   * In the site's main directory, open the _faqs folder. This is where all your FAQ posts are stored.
+
+2. **Create a New Markdown File:**
+   * Create a new file in the _faqs folder.
```
pikurasa, line 331: You should also indicate that you must append with `.md`.

As for prepending with Q1, Q2, Q3, I'm unsure if we really want to keep them numbered as they are now. We may need to shuffle the order later. We should think more about that.

pikurasa: For example, we may want to organize by topic, if these files can include multiple questions.

README.md
```diff
@@ -319,4 +319,26 @@ In this section, we will guide you through the steps to add new icons. By follow
...
+
+   ```
+   ---
+   question: "The question you want to add"
+   answer: "The answer to that question"
+   ---
+   ```
+
+4. **Save and Preview:**
+   * Save the file, then run `jekyll serve` in your terminal to start the local server. Visit `http://localhost:4000` to preview your FAQ.
```
pikurasa, line 344: I think we want to point to the instructions in the README file that indicate how to run and test a server.: https://github.com/sugarlabs/www?tab=readme-ov-file#running

That way, if our instructions change, we don't need to change them in two places.

What we may want to do, however, is update the "Running" section to indicate that `jekyll serve` (without incremental" is a good way to force new post pages in FAQ, blog, news, and events to build.

pikurasa (changes requested):
Please address my review feedback.

README.md
haroon10725: We have to change our README file. As there is no command of "jekyll serve" mentioned in it.

pikurasa: You can add to our readme something to the effect of:
`jekyll serve`: Rebuilds the entire website from scratch each time.
`bundle exec jekyll serve --incremental`: Only regenerates changed files since the last build, making development faster and more efficient.

PR #557, opened 2024-11-21
Reverting `color.adjust()` to `darken()`

mohanamisra:
Attempted to fix the sass errors due to a deprecated `darken()` function.  

However, it caused build errors and some issues with the live site.  

Hence, reverted those changes.


pikurasa:
Thank you for this, but it is now complete with https://github.com/sugarlabs/www/commit/72c28a5bb807cf7bd74b1f0d50a50f9536452557 and the website seems to be deploying properly again.

mohanamisra:
That's good to hear !

PR #586, opened 2024-12-13
make a better banner fixes#579

bhagyashree980:
Updated the _includes/nav.html file to include:
A new banner for larger screens (banner.html).
A responsive banner for smaller screens (banner-mobile.html).

### DIFF
--- a/_includes/nav.html
+++ b/_includes/nav.html
@@ -6,9 +6,15 @@
         <a class="navbar-brand" href="{{ site.baseurl }}/">
           <img src="{{ site.baseurl }}/assets/logoOpt.png" alt="Sugar Labs logo" height="90" width="270">
         </a>
-	<div>
-	  {% include banner.html %}
-	</div>
+	  <div>
+	  
+	</div><div class="fundraiser-banner desktop-banner">
+    <p>
+        <strong>Help us reach our goal of $2K this fall!</strong> 
+        Every contribution counts. Donate $3 or more to support our mission. 
+        <a href="https://paypal.com/donate-link" target="_blank" class="btn btn-primary">Donate Now</a>
+    </p>
+</div>
         <div class="d-flex justify-content-center">
           <div class="zeroMarPadBtm ">
             <ul class="mr-5  navbar-nav" id="menuBar">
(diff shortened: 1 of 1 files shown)


_includes/nav.html
```diff
@@ -6,9 +6,15 @@
         <a class="navbar-brand" href="{{ site.baseurl }}/">
           <img src="{{ site.baseurl }}/assets/logoOpt.png" alt="Sugar Labs logo" height="90" width="270">
         </a>
-	<div>
-	  {% include banner.html %}
```
pikurasa, line 10: The reason I called from a separate html is because most of the year this will be off (commented out or deleted).

pikurasa (changes requested):
Please see my feedback in the PR

quozl:
Commit message does not follow our [guidance](https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#making-commits).

_includes/nav.html
```diff
@@ -113,7 +119,14 @@
     >
       <span class="navbar-toggler-icon"></span>
     </button>
-    {% include banner-mobile.html %}
+    <div class="fundraiser-banner mobile-banner">
+    <p>
+        <strong>Support Us!</strong> 
+        Donate $3 or more to help reach our $2K goal. 
+        <a href="https://paypal.com/donate-link" target="_blank" class="btn btn-primary">Donate</a>
```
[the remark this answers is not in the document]
bhagyashree980: @pikurasa ok i change thelink please check

_includes/nav.html
bhagyashree980: @pikurasa i review this file please sir can you briefly discribe about this change i will do this for sure

bhagyashree980: @pikurasa Thank you for the clarification. I understand now that banner.html is often commented out or deleted. Would you prefer that I revert this part of the change and keep banner.html as a separate file? Please let me know how you'd like me to proceed  i will surely do this.

quozl:
I'm confused as to your intent with this pull request.  I agree with @pikurasa that separate files is reasonable so that it can be a time-based campaign.  It could also be time based by checking the date.

Your changes appear to inline the banner files, yet you haven't removed the banner files.  That's unclear; either you forgot to remove them, or you aren't convinced they should be removed.

Also, now that _master_ branch has moved, there are conflicts.  Resolve them using git and then force push, please don't create a new pull request to work around this.

quozl:
Also, please rewrite the commit message as requested.  You've changed the pull request title instead.

bhagyashree980:
@quozl ,sorry for my inconvenience actually i am so new for github that's why i make so many silly silly mistake like for commit 
and i changed the nav.html code but that not required and want their i have to change  individual files of banner.html  i am trying to solve this issue

PR #590, opened 2024-12-13
feat: Update Activity Section

soumyaranjan-panda:
https://github.com/user-attachments/assets/d6dbaa28-9ea5-451d-8c81-46bc486e1008

### DIFF
--- a/css/airspace.css
+++ b/css/airspace.css
@@ -220,32 +220,6 @@ font header .navbar-default .navbar-nav li a:hover {
   padding-top: 20px;
 }
 /*--
-  service Start
---*/
-#service {
-  text-align: center;
-  padding: 90px 0;
-}
-#service .service-item {
-  padding-bottom: 30px;
-}
-#service .service-item i {
-  font-size: 60px;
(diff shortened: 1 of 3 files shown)


quozl:
Thanks.

Please follow [our guidance for commit messages](https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#making-commits).  It isn't enough to say _update_ and what section you are updating.

Please explain why you are changing airspace.css, which is not ours to change; it is supposed to be an upstream file from another project.  When we minimise changes to upstream files it makes it much easier for us to upgrade to their later releases.

You haven't taken due care with the language of content change, and perhaps you used an LLM instead of thinking about it.  For example StartChart has replaced StarChart.  Please ensure the language is written for our elementary education target audience; parents, teachers, and especially children.  Use simpler words, as we have done.

soumyaranjan-panda:
Thank you for the feedback. I will ensure that future commit messages are more descriptive. I generally use LLMs to generate better points and simpler, easy-to-understand sentences, but I will be more careful using them further.

PR #598, opened 2024-12-20
solve the alignment issue of icon in contact us page

MOHIT-IITP:
Fixes: Solved the alignment issue of the icons in contact us Page, Minor changes. 
Before: 

After:


quozl:
Please don't change airspace.css unless you are doing so in order to bring it closer to the upstream version.

Instead, add your change as a cascaded style in custom.css

MOHIT-IITP:
@quozl  Please check again, I have updated it according to what you said.

PR #600, opened 2024-12-21
Updating the number animation of the home page

omsuneri:
I introduced the counter animation number with commas 
Before:

https://github.com/user-attachments/assets/597a5f4f-480f-43f6-853d-fb7e0cbfa87f

After:
https://github.com/user-attachments/assets/42a3bc73-c615-43ee-8f1a-31ca702fc6aa

### DIFF
--- a/js/custom.js
+++ b/js/custom.js
@@ -17,34 +17,28 @@ function toggleAnswer(answerId, element) {
 // Counter Animation of the home page 
 document.addEventListener('DOMContentLoaded', () => {
     const counters = document.querySelectorAll('.count');
-    
     // Function to animate the counter
     function animateCounter(counter) {
         const target = +counter.getAttribute('data-target');
         const speed = 200; 
 
         const updateCount = () => {
-            const current = +counter.innerText;
-            const increment = target / speed;
+            const current = +counter.innerText.replace(/,/g, '');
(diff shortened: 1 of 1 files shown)


quozl:
Please rewrite your commit messages.  https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#making-commits has our guidance.  You may need to _explain_ what you are trying to use a video for, as video does not form part of the commit message.

omsuneri:
@quozl actually earlier it was not displaying number with commas and this thing is updated by this pr well I ll update the commit message as per guidelines

omsuneri:
@quozl i had updated the PR comment take a look over it actually i and @pikurasa discussed about this issue in last sugar web meet

quozl:
Sorry, I'm not usually available for meetings unless an agenda and time is published.

Thanks for the updates to the pull request comment, however please rewrite your commit message.  `git commit --amend` followed by `git push --force`.

Now that I know your change is to add commas, how are you handling the situation for cultures where commas are not used?

pikurasa:
> Now that I know your change is to add commas, how are you handling the situation for cultures where commas are not used?

In general, we need to think about how to handle other languages and i18n. I think that, however, is outside the scope of this one issue.

PR #604, opened 2024-12-22
Fixes: Solved the hidden content by navbar in desktop mode.

Abhi-Jeeet:
Fixes :- Solved the hidden content by navbar in desktop mode.

Added padding to the slider and section title to reveal the content that was previously hidden on the webpage.

### DIFF
--- a/css/custom.css
+++ b/css/custom.css
@@ -72,9 +72,14 @@
 #social-media-logos-section .social-media-logos-img {
   padding-left: 44%;
 }
-
 .donation-bar {
   background-color: white;
   margin-top: 90px;
   padding: 10px 0; 
+}
+#slider {
+  padding-top: 20%;
+}
+.section-title1 {
+  padding-top: 10%;
 }


css/custom.css
```diff
@@ -78,3 +78,10 @@
   margin-top: 90px;
   padding: 10px 0; 
 }
+
+#slider{
+  padding-top: 20%;
+}
+.section-title1{
+  padding-top: 10%;
+}
```
quozl, line 87: Please don't forget to press enter at the end of the file.

css/custom.css
```diff
@@ -78,3 +78,10 @@
   margin-top: 90px;
   padding: 10px 0; 
 }
+
+#slider{
```
quozl, line 82: Other parts of this file have a space before `{`, why is this change different?

quozl (changes requested):
Please rewrite your commit message according to [our guidance](https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#making-commits).

i.e. don't use screenshots to explain the change, as these don't form part of the retained history.

quozl:
Thanks.

* [ ] please squash the commits; we don't need to merge three of them, and you shouldn't expect us to squash, because if we have to think about that we'll take longer,
* [ ] please follow [our guidance for writing commit messages](https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#making-commits), in particular;
  * [ ] use a short first line, a blank line, and further information below (you will have seen how GitHub breaks your commit messages, this is because your message violates convention),
  * [ ] use imperative mood, 
* [ ] please review the other pull request https://github.com/sugarlabs/www/pull/607 that is said to change the overlap, to explain if your solution is more appropriate than the change proposed in that pull request.

If you're not sure how to review, please read [our guidance for reviewing](https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#guide-for-reviewers).

quozl:
We appear to have a solution for the problem now.  Thanks for your efforts.

PR #605, opened 2024-12-22
Float icon back to top

Devmoni:
### Description

This PR updates the visibility behavior of the 'Back to Top' button. Instead of being only in the footer, the button now appears when the user scrolls down the page by 30%. This change improves the user experience by providing quicker access to the top of the page.

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -1,6 +1,12 @@
 <!-- Footer -->
 <footer  style="background-color:#282828;padding-top:5px;">
-    <h5 style="text-align:right;margin-right:20px;"><a href="#top" class="hrefCustomColor">Back to top&nbsp<i class="icon ion-arrow-up-c"></i></a></h5>
+    <h5 style="text-align:right;margin-right:20px;">
+        <a href="#top" class="hrefCustomColor" id="scrollToTopBtn" style="display:none;">
+            <svg width="24" height="15" viewBox="0 0 32 32" xmlns="http://www.w3.org/2000/svg" class="icon">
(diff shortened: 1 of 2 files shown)


quozl:
Is this feature also available in later versions of Bootstrap and Airspace?  If so would it be better to upgrade?

Devmoni:
This feature is implemented using only JavaScript and simple CSS, ensuring that it does not alter any core CSS of Bootstrap and Airspace.

I believe this can be useful since the site is quite long, helping users reach the top with just one click.

quozl:
You didn't answer my question.  I've checked, and I can find nothing in either Bootstrap or Airspace latest releases to indicate that the feature is provided by either library.

quozl:
Thanks.  My questions were about implementation and retaining our upgrade paths.  I am surprised this feature isn't in the libraries or themes we are using, but we are behind by a fair bit.

You do keep advocating afresh for your feature.  These can go into the commit message rather than the pull request.

(For myself, I find return to top buttons visually distracting and waste horizontal space at the lower edge of the view port.  I already know how to scroll to top, and don't need no web site to give me again a feature the browser already has.  Sure, there are people who don't know how to run their browser, but we're not an online gambling site.  We depend on prior learning to deliver the message that learning is critical for those in the care of the reader.).

@pikurasa please review; is this feature something you want?

quozl:
Please resolve conflicts.

Devmoni:
I have resolved the conflicts as requested.

quozl:
https://github.com/sugarlabs/www/pull/612 does replace or augment this pull request?  If replace, please close this one.  Thanks.

Devmoni:
No, as I separated #612 since it’s just a suggestion for the existing banner.

pikurasa (approved):
Tested. Looks good.

PR #611, opened 2024-12-25
Enhance Text Spacing and Adjust Donate Button for Small Screens

MOHIT-IITP:
Fixes: I adjusted the line height of the landing page's main text (p tag) to improve readability and prevent it from spreading too much on smaller screens. Additionally, I optimized the donate button to ensure it fits better on smaller screens.

### DIFF
--- a/css/custom.css
+++ b/css/custom.css
@@ -217,3 +217,18 @@
   }
   
 }
+
+@media (max-width: 767px) {
+  #slider p {
+    line-height: 1.4rem !important;
+    font-weight:800;
+
+  }
+  .donate-button{
+    height:40px;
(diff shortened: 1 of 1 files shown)


quozl:
Browser default line heights are more readable because they have been chosen to be so by font designers, and changing line heights does not make it easier to view, let alone read.  What is the line height set by theme or elements before your change?

I've also tested with Firefox → Tools → Browser Tools → Responsive Design Mode and don't see a problem with the donate button fitting on the screen.

Your commit message does not follow git convention or our project guidance.  GitHub splits it in the middle of a word; see how in their render of 7b366e8.

PR #612, opened 2024-12-26
Float Donation Button to Enhance Banner Visibility

Devmoni:
### Description
A floating button for the donation banner to ensure visibility when the current banner is not visible or partially obscured.

### Purpose
SugarLabs aims to raise $25K by this fall, and the current donation banner is visually effective. However, if a user closes the banner or scrolls down, the banner becomes hidden. This is particularly noticeable on mobile devices, where the site's "Donate" button is located within the hamburger menu, making it less accessible.

To address this, I have implemented a floating donation button. This button becomes visible when the user scrolls down and the current banner is no longer in view or when the user closes the banner. It ensures a persistent and accessible way to encourage donations at all times.

### Screenshot

image

### DIFF
--- a/_includes/donation_banner.html
+++ b/_includes/donation_banner.html
@@ -20,6 +20,71 @@
             <button class="close-button" id="close-icon">
                 <img src="../assets/entry-stop.svg" alt="Close Banner" />
             </button>
-        </div>
+      </div>
     </div>
-</div>
+  </div>
+
+  <div id="donate-popup-banner">
+    <div class="donate-container">
+        <img aria-hidden="true" class="donate-icon" src="../assets/sparkles.svg" alt="Donation Icon">
+      <div class="donate-text-container">
+        <p class="donate-text">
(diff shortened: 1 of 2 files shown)


quozl:
Reviewed.  Does seem a lot of complexity for something so simple, but I guess we'll always have a donate button.

Acceptable commit message, as if you lacked the time to write anything better, but then you go and write a huge pull request opening comment; much of that text really should be in the commit message.

Devmoni:
Hopefully, the new commit message meets the expectations :)

quozl:
Yes, thanks.

Devmoni:
@pikurasa 
I have resolved the conflict issues.Please review my PR.
Thanks

pikurasa:
I think it's a good idea, I tested it, and thanks for putting custom css in its proper place.

Devmoni:
Glad it worked :)

PR #617, opened 2024-12-29
Fix favicon "not found" error on non-homepage pages

### LINKED ISSUE #507: Favicon gets "not found" error on non-homepage pages
```
[2024-10-20 13:20:37] ERROR `/press/assets/favicon_06.png' not found.
[2024-10-20 13:21:05] ERROR `/about-us/assets/favicon_06.png' not found.
```

Screenshot from 2024-10-20 13-27-15

I've looked at our logic on /js/main.js, and I don't understand why it would be happening. It looks to me like `https://www.sugarlabs.org/assets/favicon_` is hardcoded in (Line 14).

That said, it's not clear to me whether this has any impact on the live site. If not, then perhaps we add a comment noting as much somewhere in the code.

vidipsingh:
This PR addresses the issue where favicons were not found on non-homepage pages, particularly on Firefox browser. The problem was traced to an incorrect favicon path that was not being handled properly across different pages.

### Changes Made:
- Updated the favicon link in `_layouts/base.html` to use a consistent, absolute path.
- Modified `js/main.js` to dynamically update the favicon based on a random color scheme.
- Implemented error handling in the JavaScript to revert to a default favicon if the specific one is not found.

### Issue Link:
Closes #507

### How to Test:
1. Load the homepage and any non-homepage page (e.g., `/about-us` or `/press`) on Firefox.
2. Check if the favicon is displayed correctly across all pages.
3. Ensure that a fallback favicon is shown if the dynamic one is missing.

### Additional Notes:
- This issue was specifically observed in the Firefox browser.
- The fix ensures consistent favicon display on all pages, improving the site's visual consistency and user experience.

### OS and Browser Information:
- **Operating System**: Linux
- **Browser**: Firefox


quozl:
You have a commit to add author name display to blog, yet this pull request is for the favicon problem.  Please remove the commit 9c1b988.

Also, you have placed in the pull request opening comment text that should be in the commit message of 53cb483.  Please do read our guide to making commits carefully.

vidipsingh:
@quozl, I have removed the commit `9c1b988` and will make sure to follow the guide for future contributions. 
Thank you for the feedback!

_layouts/base.html
```diff
@@ -4,7 +4,7 @@
         <meta charset="utf-8">
         <meta name="viewport" content="width=device-width, initial-scale=1">
         <meta http-equiv="X-UA-Compatible" content="IE=edge">
- 		<link id="defaultIcon1" rel="shortcut icon" href="{{ site.baseurl }}assets/favicon_06.png" />
+        <link id="defaultIcon1" rel="shortcut icon" href="/assets/favicon.png" />
```
quozl, line 7: Why is site.baseurl removed?

Why is there no file assets/favicon.png yet you link to it here?

js/main.js
```diff
@@ -16,7 +16,24 @@ $(document).ready(function () {
...
+        var faviconPath = '/assets/favicon_' + logoID + '.png';
+        
+        // Test if favicon exists before setting
+        var img = new Image();
+        img.onload = function() {
+            defaultIcon.href = faviconPath;
+        };
+        img.onerror = function() {
+            // Fallback to default favicon
+            defaultIcon.href = '/assets/favicon.png';
```
quozl, line 32: There is no file assets/favicon.png

js/main.js
```diff
@@ -16,7 +16,24 @@ $(document).ready(function () {
...
         logoID = "0" + logoID;
     }
-    document.querySelector('#defaultIcon1').href = 'https://www.sugarlabs.org/assets/favicon_' + logoID + '.png';
+
+    var defaultIcon = document.querySelector('#defaultIcon1');
+    if (defaultIcon) {
+        // Use absolute path for favicon
+        var faviconPath = '/assets/favicon_' + logoID + '.png';
+        
+        // Test if favicon exists before setting
```
quozl, line 25: I don't see why we need to test if it exists.  We know exactly how many files we have, what their names are, and we know the code above this section in js/main.js.

js/main.js
```diff
@@ -16,7 +16,24 @@ $(document).ready(function () {
     if (logoID < 10) {
         logoID = "0" + logoID;
     }
-    document.querySelector('#defaultIcon1').href = 'https://www.sugarlabs.org/assets/favicon_' + logoID + '.png';
```
quozl, line 19: Code you are removing has nothing to do with the issue; the issue and my tests of master branch just now show the error to be a reference to press/assets/favicon_06.png and this line you are removing is an absolute link without press.

quozl (changes requested):
I'm not convinced this is the right approach.  Details are in the feedback below.

* [ ] rewrite commit messages as suggested earlier.

vidipsingh:
> I'm not convinced this is the right approach. Details are in the feedback below.
> 
> * [ ]  rewrite commit messages as suggested earlier.

Hi @quozl,

Thanks for the feedback. I’ll work on the changes as suggested. 
Could you please point out what specifically is wrong with my current commit message? It would help me improve from next time.

_layouts/base.html
vidipsingh: I will revert the changes in `_layouts/base.html` to the previous version. 
Since there is no `/favicon.png`, should I refer to `/assets/favicon_06.png` here instead as before?

js/main.js
vidipsingh: Since there is no `/assets/favicon.png`, should I refer to `/assets/favicon_06.png` here instead?

js/main.js
vidipsingh: Sure, I'll remove the favicon test since we already know the number and names of the images.

js/main.js
vidipsingh: I have simplified the favicon logic by making the path relative to `/press/`, removing the full domain URL, and eliminating unnecessary existence checks, as shown in the following code:

```javascript
var defaultIcon = document.querySelector('#defaultIcon1');
if (defaultIcon) {
    var logoID = colorIndex + 1;
    if (logoID < 10) {
        logoID = "0" + logoID;
    }
    defaultIcon.href = '/press/assets/favicon_' + logoID + '.png';
}
```
Also, I have removed the testing and existence checks. Is this approach fine?

quozl:
Sure.  When I went to your commit message and read it, assuming the persona of a future developer trying to figure out why `git bisect` identified the commit as the cause of a problem being investigated, I could not find the information we expect;

* [ ] explain the problem that is solved; you've put this in the pull request opening comment instead of the commit message,
* [ ] when the problem was introduced by a previous commit, mention the hash; this shows your full understanding and helps with review, letting us see how the problem began and evaluate if your solution is the right one,
* [ ] when the problem is in an issue or ticket, as it does, end with "Fixes #507".

It is alarming when a pull request opening comment has such a huge amount of detail that is left out of the commit message.

Remember that the git commit message is kept in perpetuity, and pull requests are deleted eventually.  The git repository clone on a developer's system is where everything necessary to development must be present.  GitHub is only a public hosting point for the repository.

_layouts/base.html
quozl: Don't mind either way, just that it should refer to something that exists.

js/main.js
quozl: As above.

js/main.js
quozl: You've lost me here.  Why would the path be made absolute from /press/?

_layouts/base.html
vidipsingh: Sure, then I will change it to `{{ site.baseurl }}/assets/favicon_06.png`.

js/main.js
vidipsingh: Ok sure.

js/main.js
vidipsingh: Sorry, I think relative path will be better for this like:
```javascript
var defaultIcon = document.querySelector('#defaultIcon1');
if (defaultIcon) {
    var logoID = colorIndex + 1;
    if (logoID < 10) {
        logoID = "0" + logoID;
    }
    defaultIcon.href = 'assets/favicon_' + logoID + '.png';
}
```
Now it's pointing to the `assets/` directory relative to the current location, which should be more flexible.

vidipsingh:
> Sure. When I went to your commit message and read it, assuming the persona of a future developer trying to figure out why `git bisect` identified the commit as the cause of a problem being investigated, I could not find the information we expect;
> 
> * [ ]  explain the problem that is solved; you've put this in the pull request opening comment instead of the commit message,
> * [ ]  when the problem was introduced by a previous commit, mention the hash; this shows your full understanding and helps with review, letting us see how the problem began and evaluate if your solution is the right one,
> * [ ]  when the problem is in an issue or ticket, as it does, end with "`Fixes #507`", so that GitHub will automatically close the issue.
> 
> It is alarming when a pull request opening comment has such a huge amount of detail that is left out of the commit message.
> 
> Remember that the git commit message is kept in perpetuity, and pull requests are deleted eventually. The git repository clone on a developer's system is where everything necessary to development must be present. GitHub is only a public hosting point for the repository.

Thanks for the feedback, @quozl! I’ll make sure to incorporate these practices in future commits and ensure all relevant details are included directly in the commit message.

quozl:
> Thanks for the feedback, @quozl! I’ll make sure to incorporate these practices in future commits and ensure all relevant details are included directly in the commit message.

So you're saying you won't fix it for this pull request?  Why not?  Our purpose for review includes managing the history.

You have just one commit.  All you have to do is `git commit --amend` followed by `git push --force`.  If you had more than one commit, you would `git rebase -i`, select _reword_, then `git push --force`.

See our [guide for reviewers](https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#guide-for-reviewers) to understand our review practices, and please learn to review your own work and others.  We need more reviewers.

vidipsingh:
> > Thanks for the feedback, @quozl! I’ll make sure to incorporate these practices in future commits and ensure all relevant details are included directly in the commit message.
> 
> So you're saying you won't fix it for this pull request? Why not? Our purpose for review includes managing the history.
> 
> You have just one commit. All you have to do is `git commit --amend` followed by `git push --force`. If you had more than one commit, you would `git rebase -i`, select _reword_, then `git push --force`.
> 
> See our [guide for reviewers](https://github.com/sugarlabs/sugar-docs/blob/master/src/contributing.md#guide-for-reviewers) to understand our review practices, and please learn to review your own work and others. We need more reviewers.

Sorry about that, @quozl! I’ll make the necessary changes to the recent commit and ensure to follow these practices for all future commits as well.

quozl:
@pikurasa ready for your review; this pull request has finished technical review and is on the way toward commit message goodness, just need to know you are happy with the solution to the issue.

js/main.js
```diff
@@ -16,7 +16,16 @@ $(document).ready(function () {
...
+
+    var defaultIcon = document.querySelector('#defaultIcon1');
+    if (defaultIcon) {
+        var logoID = colorIndex + 1;
+        if (logoID < 10) {
+            logoID = "0" + logoID;
+        }
+        defaultIcon.href = 'assets/favicon_' + logoID + '.png';
```
pikurasa, line 26: Is this still taking a relative path?

I ask because I still see things like this in the logs:

Screenshot from 2025-01-04 16-25-26

quozl: Agreed, is use of relative path.  Oops.

quozl:
Noted edited commit message on https://github.com/sugarlabs/www/pull/617/commits/1939ac9f9d07b2c1a4631740b32fd6d577abc9c7

Review comments;

* [ ] the first line is too long, see how GitHub renders it by _breaking in the middle of a word_, which is them telling you to fix it,
* [ ] remove the # before the commit hash so it is properly rendered by GitHub, like this 58177ea,
* [ ] put the "fixes" and issue number at the very end.

js/main.js
vidipsingh: > Is this still taking a relative path?
> 
> I ask because I still see things like this in the logs:
> 
> Screenshot from 2025-01-04 16-25-26

Yes, it is now taking relative paths as intended. Previously, the code was referencing the favicon like this:

```javascript
document.querySelector('#defaultIcon1').href = 'https://www.sugarlabs.org/assets/favicon_' + logoID + '.png';
```

However, I have updated it to use relative paths, as shown here:
```javascript
defaultIcon.href = 'assets/favicon_' + logoID + '.png';
```


After this change, I am no longer seeing any errors in my logs, whether on Firefox or other browsers.

image

quozl: Well, relative or absolute, doesn't matter, as long as the directory that contains the icons is accessed on both the root and lower pages.  It works for the root page, but as @pikurasa said it doesn't work on lower pages, such as `about-us/`.  That's what you need to fix now.

quozl: (a relative access from `about-us/` would be to `../assets/`)

vidipsingh: > Well, relative or absolute, doesn't matter, as long as the directory that contains the icons is accessed on both the root and lower pages. It works for the root page, but as @pikurasa said it doesn't work on lower pages, such as `about-us/`. That's what you need to fix now.

In the `_layouts/base.html`, the previous line was:

```html
<link id="defaultIcon1" rel="shortcut icon" href="{{ site.baseurl }}assets/favicon_06.png" /> 
```

I changed it to:
```html
<link id="defaultIcon1" rel="shortcut icon" href="{{ site.baseurl }}/assets/favicon_06.png" /> 
```

The issue was that there was no forward slash between `{{ site.baseurl }}` and `assets`, which caused incorrect path resolution when navigating between pages like `/about-us/`. Now, the problem of the favicon "not found" error on non-homepage pages is not there in Firefox.

vidipsingh: > Yes, I know, but I was not talking about that part of the change. I was talking about the part of the change that this discussion thread is linked to; js/main.js line 26, which loads a relative path into the element href, without dereferencing the current document's relative path, which is why the browser then reports failures to find the png files.


@quozl The change that I made for this is as follows:

```javascript
var defaultIcon = document.querySelector('#defaultIcon1');
if (defaultIcon) {
    var logoID = colorIndex + 1;
    if (logoID < 10) {
        logoID = "0" + logoID;
    }

    // Extract the base path from the current favicon href
    var currentHref = defaultIcon.getAttribute('href');
    var basePath = currentHref.substring(0, currentHref.lastIndexOf('/') + 1);
    
    // Update only the filename portion while keeping the original path
    defaultIcon.href = basePath + 'favicon_' + logoID + '.png';
}
```

And in `_layouts/base.html`:
```javascript
<link id="defaultIcon1" rel="shortcut icon" href="{{ site.baseurl }}/assets/favicon_06.png" />
```

This approach resolves the issue of subdirectories by dynamically extracting the base path from the current favicon and then updating only the filename. By preserving the base path that starts from `{{ site.baseurl }}`, this ensures that the favicon works across all pages, regardless of how deep they are within the site's directory structure.

Would this be a good approach to resolve the issue? I’d also appreciate any suggestions you might have to improve it.

quozl: Hope so.  Commit the change and push to the pull request branch so that I can see it in context and we can test.

quozl:
Tested 245ea8c with Firefox in a container using `--host`.

* does fix #507,
* causes double GET of favicon;
  * often the initial GET is aborted asynchronously; is there a way to avoid that, such as by making the href change inside the page rather than in external script?
  * otherwise the initial GET completes and then the js/main.js triggers another GET; is there a way to avoid two GETs?
* removing _defer_ makes no difference,
* your branch diverged from master a few weeks ago, so all the things we've since fixed come back, please keep your pull request branch up to date.

vidipsingh:
> Tested [245ea8c](https://github.com/sugarlabs/www/commit/245ea8cdae5f20236727dcb2adca67b004d38d85) with Firefox in a container using `--host`.
> 
> * does fix [Favicon gets "not found" error on non-homepage pages #507](https://github.com/sugarlabs/www/issues/507),
> * causes double GET of favicon;
>   
>   * often the initial GET is aborted asynchronously; is there a way to avoid that, such as by making the href change inside the page rather than in external script?
>   * otherwise the initial GET completes and then the js/main.js triggers another GET; is there a way to avoid two GETs?
> * removing _defer_ makes no difference,
> * your branch diverged from master a few weeks ago, so all the things we've since fixed come back, please keep your pull request branch up to date.

Hi @quozl,
Thanks for your feedback.
- I'm also running the setup with Firefox in a container using `--host`, but I haven’t been able to observe the double GET request for the favicon in the console or elsewhere. Could you kindly elaborate on how and where exactly the two GET requests for the favicon are being triggered? It would be very helpful if you can provide more details or steps to reproduce the issue.
- As for the `defer` attribute, since removing it doesn't seem to make any difference in the site's functionality or error logging, should I go ahead and remove it?
- I'll make sure to update my branch with the latest changes from `master`.

quozl:
> I'm also running the setup with Firefox in a container using `--host`, but I haven’t been able to observe the double GET request for the favicon in the console or elsewhere.

`jekyll` does not appear to have an option to provide a server side request log.  It is possible to use `tcpdump` or `wireshark` but takes longer to analyse.

> Could you kindly elaborate on how and where exactly the two GET requests for the favicon are being triggered? It would be very helpful if you can provide more details or steps to reproduce the issue.

Sure.  Method used;
* Start Jekyll,
* Start Firefox and open the site using the URL provided by Jekyll,
* Tools → Browser Tools → Web Developer Tools,
* Network tab,
* Reload,
* Filter "favicon".

Observed one of three scenarios, which is unsurprising given this is a race condition between the browser's built-in loading of favicon and our script in js/main.js;

**Scenario 0 - browser begins GET but cancels due to script**

fi0

**Scenario 1 - browser GET followed by script GET**

fi1

**Scenario 2 - browser GET but the script chooses the same random icon**

fi2

> As for the `defer` attribute, since removing it doesn't seem to make any difference in the site's functionality or error logging, should I go ahead and remove it?

Why would you do that?

vidipsingh:
> > I'm also running the setup with Firefox in a container using `--host`, but I haven’t been able to observe the double GET request for the favicon in the console or elsewhere.
> 
> `jekyll` does not appear to have an option to provide a server side request log. It is possible to use `tcpdump` or `wireshark` but takes longer to analyse.
> 
> > Could you kindly elaborate on how and where exactly the two GET requests for the favicon are being triggered? It would be very helpful if you can provide more details or steps to reproduce the issue.
> 
> Sure. Method used;
> 
> * Start Jekyll,
> * Start Firefox and open the site using the URL provided by Jekyll,
> * Tools → Browser Tools → Web Developer Tools,
> * Network tab,
> * Reload,
> * Filter "favicon".
> 
> Observed one of three scenarios, which is unsurprising given this is a race condition between the browser's built-in loading of favicon and our script in js/main.js;
> 
> **Scenario 0 - browser begins GET but cancels due to script**
> 
> fi0
> 
> **Scenario 1 - browser GET followed by script GET**
> 
> fi1
> 
> **Scenario 2 - browser GET but the script chooses the same random icon**
> 
> fi2
> 
> > As for the `defer` attribute, since removing it doesn't seem to make any difference in the site's functionality or error logging, should I go ahead and remove it?
> 
> Why would you do that?

I think the problem is:
1. When the browser loads the page, it first detects the default favicon in the HTML:

```html
<link id="defaultIcon1" rel="shortcut icon" href="{{ site.baseurl }}/assets/favicon_06.png" />
```

2. The JavaScript in `main.js` then attempts to randomly change the `favicon`:

```javascript
var defaultIcon = document.querySelector('#defaultIcon1');
if (defaultIcon) {
    var logoID = colorIndex + 1;
    if (logoID < 10) {
        logoID = "0" + logoID;
    }
    // Update favicon path
    defaultIcon.href = basePath + 'favicon_' + logoID + '.png';
}
```
This creates a _race condition_ which is given in the above three scenarios.

One solution I can think of is that we can:
- Moving `favicon` generation to before the DOM content loads
- Removing the static `favicon` from HTML
- Dynamically creating and injecting the `favicon` element using JavaScript
- Using a `meta` tag to store the base path

However, the issue only appears during the first request, and subsequent `GET` requests for the favicon work fine in Firefox for me.

As for the `defer`attribute I think it should be there only as it is helpful because it loads the script in the background without blocking the parsing of the HTML, and only executes it when the DOM has been completely constructed.

quozl:
Thanks.

Removing the static favicon would probably cause the browser to attempt `favicon.ico` as the default.  Let me know what you find.

My guess is that changing the href in a page script may happen sooner than js/main.js

Yes, the issue is only there on first request of a page, but our target environment has been users with bad internet.  I'd like to minimse off-site and repeated GETs.

We have Google Analytics loading as well, but if nobody is using the results we don't need to do that.

vidipsingh:
Hi @quozl,

Referring to the favicon issue fix, it’s now working perfectly in Firefox! The favicon loads correctly on sub-URLs, ensuring the icon is displayed consistently across different page depths.

Here’s the full updated code in `main.js`:
```javascript
var defaultIcon = document.querySelector('#defaultIcon1');
    if (defaultIcon) {
        var logoID = colorIndex + 1;
        if (logoID < 10) {
            logoID = "0" + logoID;
        }
        
        // Get the site's root URL by looking at the <base> tag or constructing from window.location
        var rootUrl = '';
        var baseTag = document.querySelector('base');
        if (baseTag && baseTag.href) {
            rootUrl = baseTag.href;
        } else {
            // Construct the root URL from window.location
            rootUrl = window.location.protocol + '//' + window.location.host + '/';
        }
        
        // Use absolute path for favicon
        defaultIcon.href = rootUrl + 'assets/favicon_' + logoID + '.png';
    }
```

Please share some feedback or a review on the fix.

quozl:
As it is only a change to js/main.js, I can't see how it would change the browser default loading of what is specified in base.html

Perhaps you might push your changes to the branch so I can test them?

pikurasa:
I tested https://github.com/vidipsingh/www/commit/7c5c491da8ed15ebf1337413ac0e76f97425f7e1, and it seems to work. (Previous commits did not solve the issue for me.)

@vidipsingh @quozl is there anything else that we're waiting on here?

quozl:
Thanks for testing.

vidipsingh:
Thank you @pikurasa and @quozl for your feedback, and for merging the PR!

Issue #638, opened 2025-01-14
Better readablity and UX for numbers section


Vishvsalvi:
# Before
![{A50C4A41-1320-4FDB-9F45-697FEBE4B30A}](https://github.com/user-attachments/assets/bfef4e5d-0848-4032-843a-af3dfe5268d1)

# After

![{D4FEC879-EB8E-4F2A-A431-56747A9266BC}](https://github.com/user-attachments/assets/3301a9c5-3a10-4aee-a46d-bad72394b595)

pikurasa:
What is the issue?

PR #640, opened 2025-01-14
Better row_col arrangement

Vishvsalvi:
Solving #638

Screenshot_20250114_204420


pikurasa:
What is the improvement?

PR #649, opened 2025-01-18
Add Product page

therealharshit:
@pikurasa here is the product page as we discussed
Screenshot 2025-01-19 014634
please review.


pikurasa:
I tested it, and I'm merging it now.

I do, however, wonder if putting it under "donate" will be confusing for some people.

We'll also need to link to the USB from https://www.sugarlabs.org/booting-soas/

therealharshit:
> I tested it, and I'm merging it now.
> 
> I do, however, wonder if putting it under "donate" will be confusing for some people.
> 
> We'll also need to link to the USB from https://www.sugarlabs.org/booting-soas/

@pikurasa I thought purchasing products from sugarlabs is also a way of donating to the organization that's why I kept it under 'donate'. 

Please let me know if something needs to be changed.

quozl:
"donate" is certainly not related to "purchase".

therealharshit:
> "donate" is certainly not related to "purchase".

I have made some changes, please review it in PR #658 .

PR #652, opened 2025-01-20
fixes/#650 needless favicon fetch fix

### LINKED ISSUE #650: Needless double GET of favicon
On loading the site does GET favicon twice, once because of the link tag, and once because of js/main.fs.

Reduce to one or zero GETs.

ad1tyayadav:
This PR fixes #650 

https://github.com/user-attachments/assets/622fd912-1465-4e7c-aa48-6a7b66b25dcc

### DIFF
--- a/js/main.js
+++ b/js/main.js
@@ -12,18 +12,13 @@ $(document).ready(function () {
     var selectedColors = COLORS[colorIndex];
 
     //Apply selected colors to the Sugarlabs logo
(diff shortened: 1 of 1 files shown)


quozl (changes requested):
Doesn't fix the problem; when the icon is changed in line 33 of your code, the browser has still begun to GET the icon that was set by the include and that GET is cancelled.

Issue #637, opened 2025-01-14
Remove unnecessary slider component 


Vishvsalvi:
![Screenshot_20250114_172213](https://github.com/user-attachments/assets/ab74bced-477d-4c47-bc81-5cb6ec1d4256)

All the social media links can be displayed horizontally instead of the slider. The white space will be properly utlized

pikurasa:
Yes, this is a problem. Thanks for raising the issue.

Ofc-Pranjal:
Can I fix this?

anya-xcode:
@pikurasa  I  am working on this issue

Vishvsalvi:
I've already made a PR

pikurasa:
> [@pikurasa](https://github.com/pikurasa) I am working on this issue

FWIW, it's still a WIP.

PR #654, opened 2025-01-20
Unwanted carousel

Vishvsalvi:
# Before
{B10CE6A1-A842-48FF-833A-BE1CA66DA5AD}

# After
{FBF7A476-226A-4E78-B00D-FA9E20845006}

Fixing #637

Removed unwanted carousel

### DIFF
--- a/booting-soas.html
+++ b/booting-soas.html
@@ -34,112 +34,42 @@ permalink: /booting-soas/index.html
       <h3 class="text-center customMargin1" >FOLLOW THE STEPS BELOW...</h3>
     </div>
     <div class="row">
-      <div class="col-md-12" >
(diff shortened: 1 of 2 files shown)


pikurasa:
Did you push the right branch? This seems to have non-related commits.

It also has unrelated changes, so it's hard to read.

pikurasa:
Also, the "after" image has white on white, which is hard to see.

If that is the current state of your code, I suggest fixing that as well.

Vishvsalvi:
> Also, the "after" image has white on white, which is hard to see.
> 
> If that is the current state of your code, I suggest fixing that as well.

I'm sorry for that, I'll fix that asap

PR #658, opened 2025-01-25
Update the position of "Products" on nav

therealharshit:
Updated the position of "Products" section from "Donate" dropdown to a separate item on navbar.

Screenshot 2025-01-25 201223

mobile:
Screenshot 2025-01-25 201327

@pikurasa what do you think.

### DIFF
--- a/_includes/nav.html
+++ b/_includes/nav.html
@@ -43,13 +43,11 @@
               <li class="nav-item">
                 <a class="nav-link" href="{{ site.baseurl }}/volunteering">Volunteer</a>
               </li>
-              <li class="nav-item dropdown align-self-center">
-                <a href="{{ site.baseurl }}/#" class="dropdown-toggle" data-toggle="dropdown" role="button"
-                  aria-haspopup="true" aria-expanded="false">Donate<span class="caret"></span></a>
-                <div class="dropdown-menu">
-                  <a class="dropdown-item" href="{{ site.baseurl }}/donate">Donate</a>
-                  <a class="dropdown-item" href="{{ site.baseurl }}/product">Products</a>
-                </div>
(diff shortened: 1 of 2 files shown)


pikurasa:
I was about to ask that the URL be updated to "products" (plural), instead of "product", but I hesitate because I already shared the "product" link.

@therealharshit is there a way to forward, perhaps via https://github.com/sugarlabs/www/blob/master/.htaccess ?

Whenever we move a link, we should add a forward. If we cannot forward, then we should probably leave be and just be extra careful when naming links.

therealharshit:
> I was about to ask that the URL be updated to "products" (plural), instead of "product", but I hesitate because I already shared the "product" link.
> 
> @therealharshit is there a way to forward, perhaps via https://github.com/sugarlabs/www/blob/master/.htaccess ?
> 
> Whenever we move a link, we should add a forward. If we cannot forward, then we should probably leave be and just be extra careful when naming links.

Sorry for the mistake I did while naming the link.

I think now you want me to update the link from /product to /products.
And also want people to get redirected to /products whenever the click /product link. 

I tried to do this by https://github.com/sugarlabs/www/blob/master/.htaccess but it's not working, I think this approach is very server specific.

But I am working on a way to redirect using Javascript.

pikurasa:
> But I am working on a way to redirect using Javascript.

Don't spend too much time on finding a solution. If it's easy, let's do it. If not, let's forget about it for now.

therealharshit:
> > But I am working on a way to redirect using Javascript.
> 
> Don't spend too much time on finding a solution. If it's easy, let's do it. If not, let's forget about it for now.

@pikurasa Please review my proposed solution in #664.

PR #634, opened 2025-01-09
Replace deprecated darken with color.adjust for Sass compliance

uditjainstjis:
### Summary 
- This code fixes #633  the warning of depreciated darken function in the terminal and 
- Removes bug of a '{' bracket missing in the code.
- Watching file was made for compiling the '.scss' file.
### Changes/explanation

- darken function was replaced with color.adjust which is supported
- '{' bracket was added at the missing place
- Ran the command: `sass /css/ion_icon_customization.scss /css/ion_icon_customization.css` to generate the compiled css file, which was necessary for "@use keyword" in @use "sass:color" which was used to use color thing to replaced darken the main thing.
### Testing

- Build the site locally with jekyll serve and verify image and list rendering.
- Check bullet points and image scaling on mobile and desktop views.
- Ensured now no error is present in the terminal and is clear.

PR #525, opened 2024-10-27
delete Gemfile.lock from repository

haroon10725:
@pikurasa Can you please review it.

PR #676, opened 2025-02-02
Fixed missing dependencies issue by updating Gemfile

### LINKED ISSUE #670: Error running bundle exec jekyll serve --incremental due to missing dependencies while set up
When running bundle exec jekyll serve --incremental, the following error occurs during the set up of the project.

In mac:
```
/opt/homebrew/lib/ruby/gems/3.4.0/gems/jekyll-4.3.4/lib/jekyll.rb:26: warning: logger was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0........
```
In windows:
```
C:/Users/Hp/AppData/Local/Microsoft/WindowsApps/jekyll:25: warning: csv was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.4.0.
You can add csv to your Gemfile or gemspec to silence this warning........
```

sa-fw-an:
When running bundle exec jekyll serve --incremental, an error occured. Investigation reveals that the issue is due to missing dependencies in the Gemfile, which was introduced by Commit `1dd0e41.` The changes in this commit removed essential dependencies, leading to this error.

- Fixes #670

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,12 +1,26 @@
 source "https://rubygems.org"
-gem "jekyll", "~> 4.3.4"
-gem "minima", "~> 2.5"
+
+# Core Jekyll and theme
+gem "jekyll"
+gem "minima"
+
+# Jekyll plugins
 group :jekyll_plugins do
-  gem "jekyll-feed", "~> 0.12"
+  gem "jekyll-feed"
+  gem "github-pages"
+  gem "jekyll-sitemap"
 end
+
+# Platform-specific gems
 platforms :mingw, :x64_mingw, :mswin, :jruby do
-  gem "tzinfo", ">= 1", "< 3"
+  gem "tzinfo"
   gem "tzinfo-data"
 end
-gem "wdm", "~> 0.1", :platforms => [:mingw, :x64_mingw, :mswin]
-gem "http_parser.rb", "~> 0.6.0", :platforms => [:jruby]
+
+gem "wdm", platforms: [:mingw, :x64_mingw, :mswin]
+gem "http_parser.rb", platforms: [:jruby]
+
+# Additional gems
+gem "bundler"
+gem "webrick"
+gem "rake"


sa-fw-an:
@pikurasa Ive created a new pr. Please review this

quozl:
That's a future deprecation warning, not an error.  Issue closed.

You're apparently updating the dependencies using one of the steps mentioned in 
https://github.com/sugarlabs/www/pull/634#issuecomment-2608849733

But without the other steps, this change would likely break the web site deployment build.

sa-fw-an:
Hi @quozl,
I've attached a screenshot of the error that occurs when running `bundle exec jekyll serve --incremental`
#### Error Screenshot:
----
Screenshot 2025-02-03 at 7 19 38 AM
--
After investigating, I found that the issue was caused by missing dependencies. To resolve this, I've raised a PR that fixes the problem. Please review it.
To troubleshoot, could you try setting up the dev environment again after removing all locally installed gems? This will help verify if the error persists.
Others have encountered the same issue, and it's likely that PR
- #525 

introduced the problem. I look forward to your response.

Best regards.

quozl:
As I said, we're using Jekyll 3, and moving to Jekyll 4 as you have suggested will be a complex process that risks the build and deploy of the web site.  I don't see any sign that you know what you are talking about in the context of that process.

pikurasa:
I've reviewed the PR, and I don't quite see where this PR would move us to Jekyll 4.

While we might not know for certain if it would work on GH if merged, FWIW it does work on my machine.

Also, in some ways it seems like a simplification, since it is removing constraints on which versions to load. So I assume that whatever setup you have, your system resolves the dependencies on its own.

So, I'm not totally certain yet that we write this PR off.

That said, I never encountered the issue. So it's value is mainly in helping contributors using Mac/Windows up and running with fewer problems, and it doesn't solve any pressing issues with deployment.

sa-fw-an:
> I've reviewed the PR, and I don't quite see where this PR would move us to Jekyll 4.
> 
> While we might not know for certain if it would work on GH if merged, FWIW it does work on my machine.
> 
> Also, in some ways it seems like a simplification, since it is removing constraints on which versions to load. So I assume that whatever setup you have, your system resolves the dependencies on its own.
> 
> So, I'm not totally certain yet that we write this PR off.
> 
> That said, I never encountered the issue. So it's value is mainly in helping contributors using Mac/Windows up and running with fewer problems, and it doesn't solve any pressing issues with deployment.

@pikurasa Yes they would automatically solve the dependency errors. Also new begginers wouldn't be facing issue they are facing now if this is merged. Can i go ahead with a new pr ?

quozl:
I've not yet found a good reason to move to Jekyll 4.  New beginners could use the Docker set up by @MostlyKIGuess, but the biggest mistake seems to be their choice of task; mostly cosmetic.  Changes to content in response to changes in the environment, code, or use of Sugarizer, Music Blocks, or Sugar would be more welcome.

You can have more than one pull request open at a time, that's your choice.  Just don't mix the branches and be ready to rebase when _master_ branch moves.  It's a skill.

PR #678, opened 2025-02-16
added download links for MB

### LINKED ISSUE #677: No download Links for MB in music-blocks.html
Image

Since the page for sugarizer has all the information about all the distribution platforms sugarizer is available for download. Music Blocks page should follow with the same consistency. Therefore:

- Music Block page should have a section for all its distributors 
- It should have a clickable button which will redirect to installation pages for the respective distributors. 
- It should also maintain a level of consistency of design without adding unnecessary stylesheets.

ChinmayVS:
closes #677 
added music block download links for 
- Playstore
- Flatpak

image

### DIFF
--- a/music-blocks.html
+++ b/music-blocks.html
@@ -144,6 +144,48 @@ permalink: /music-blocks/index.html
     </div>
 </section>
 
+<!-- "Download MB" section -->
+<section id="call-to-action-mb" style="padding: 40px 0; text-align: center; background-color: #fff;">
+    <div style="max-width: 1200px; margin: 0 auto; padding: 20px;">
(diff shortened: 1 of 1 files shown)


pikurasa (changes requested):
Please make this one change.

Also, I think that under our current "Try Music Blocks" button it should have a sentence that says something like "Open Music Blocks directly in your browser. Works best on larger screens, like laptops, desktops, and larger tablets." I recommend this so that people don't mistakenly think that they must download the software in order to run it.

pikurasa:
@ChinmayVS  Please make the suggested change, and I can review it again for merging.

PR #688, opened 2025-03-25
set carousel autoPlay to false 

haroon10725:
@pikurasa Can you please review it?


pikurasa (approved):
It's good enough for now...

PR #694, opened 2025-06-21
Replace deprecated `darken()` function with `color.scale()` for Dart Sass 3+ compatibility

geekyraghav13:
This PR replaces all instances of the deprecated `darken()` function in the `ion_icon_customization.scss` file with the recommended `color.scale()` method for compatibility with Dart Sass 3.0.0 and above.

### Changes Made:
- Updated all `darken($color, $percent)` calls to `color.scale($color, $lightness: -$percent)`
- Ensured the file compiles without deprecation warnings
- Verified all nested styles and variables remain intact with the new method

This change helps eliminate terminal warning floods and ensures future Sass compatibility.


quozl:
Thanks, but not required.

PR #695, opened 2026-03-16
docs: fix typos in README

dreampretti15:
This PR fixes a couple of small typos in the README.


quozl:
Heh.  Are the typos in the new web site?  We're not using this repository any more, so don't think we really need to fix typos.

dreampretti15:
Thank you for the clarification.

I understand — since the repository is no longer in use, this change may not be necessary.

I’ll look for other issues to contribute to.

Thank you again!

